PR #3, opened 2018-12-06
Feature/update charts

jschaul:


tiago-loureiro:
The new webapp variables are not yet available on the build version included in this PR. I suggest updating the version to what is here: https://github.com/wireapp/wire-server-deploy/blob/web-upgrades/charts/webapp/values.yaml#L12-L26 , including a few extra comments on the values section.

PR #7, opened 2019-01-30
Fix external endpoints

tiago-loureiro:
Relevant changes:
 * Fixes external endpoints to be reachable from the nginx-ingress
 * Upgrades nginx-ingress to the latest version
 * Allows using an external service in the S3 part of the nginx-ingress (useful if minio is outside of k8s)
 * Added a missing ingress for team settings

The most important bit is https://github.com/wireapp/wire-server-deploy/commit/8aef810076bbed95370a3573db966edd61ff7f7c - particularly the fact that the name of the ports in the `Endpoints` must match the `Service` in order for this to work - I am not sure how/why this worked before: it worked "internally" (i.e., brig was able to reach an external cassandra) but the nginx-ingress was not, kept returning a 503

Relevant references:

https://appscode.com/products/voyager/7.1.1/guides/ingress/http/external-svc/#using-external-ip
https://github.com/kubernetes/kubernetes/issues/8631#issuecomment-104760910
https://docs.openshift.com/enterprise/3.0/dev_guide/integrating_external_services.html - _The port and name definition must match the port and name value in the service_

The [publishNotReadyAddresses](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/) may be interesting too for peer discovery purposes.

### DIFF
--- a/charts/cassandra-external/templates/endpoint.yaml
+++ b/charts/cassandra-external/templates/endpoint.yaml
@@ -32,5 +32,7 @@ subsets:
       - ip: {{ . }}
       {{- end }}
     ports:
+      # port and name in the endpoint must match port and name in the service
+      # see also https://docs.openshift.com/enterprise/3.0/dev_guide/integrating_external_services.html
       - name: cql
         port: {{ .Values.portCql }}
--- a/charts/cassandra-external/templates/endpoint.yaml
+++ b/charts/cassandra-external/templates/endpoint.yaml
@@ -32,4 +32,5 @@ subsets:
       - ip: {{ . }}
       {{- end }}
     ports:
-      - port: {{ .Values.portCql }}
+      - name: cql
+        port: {{ .Values.portCql }}
--- a/charts/minio-external/templates/endpoint.yaml
+++ b/charts/minio-external/templates/endpoint.yaml
@@ -32,5 +32,7 @@ subsets:
       - ip: {{ . }}
       {{- end }}
     ports:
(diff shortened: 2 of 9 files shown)


charts/cassandra-external/templates/endpoint.yaml
```diff
@@ -32,4 +32,5 @@ subsets:
       - ip: {{ . }}
       {{- end }}
     ports:
-      - port: {{ .Values.portCql }}
+      - name: cql
```
jschaul:
```suggestion
      # port and name in the endpoint must match port and name in the service
      # see also https://docs.openshift.com/enterprise/3.0/dev_guide/integrating_external_services.html
      - name: cql
```

charts/minio-external/templates/endpoint.yaml
```diff
@@ -32,4 +32,5 @@ subsets:
       - ip: {{ . }}
       {{- end }}
     ports:
-      - port: {{ .Values.portHttp }}
+      - name: minio
```
jschaul:
```suggestion
      # port and name in the endpoint must match port and name in the service
      # see also https://docs.openshift.com/enterprise/3.0/dev_guide/integrating_external_services.html
      - name: minio
```

tiago-loureiro:
> Where exactly would this be useful?

Yeah sry, not directly related to this PR ;) This is mostly for stateful sets (e.g., C* or something else that we might want to try running on k8s) since they may not be able to provide valid readiness checks until more than n nodes have actually bootstrapped

PR #8, opened 2019-02-04
Publish latest charts and made BSD friendlier bash script

tiago-loureiro:
`sed -i` does not work really work on osx, this should work on different systems

Also, published the latest charts and make some change `wire-tmp -> $INDEX_S3_DIR` which seemed to be missing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.tgz
 *secrets.yaml
 *.key
+*.bak
 *.cert
 demo-values.yaml
 requirements.lock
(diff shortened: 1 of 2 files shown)


tiago-loureiro:
If `bin/sync.sh` fails, please have a look at the changes https://github.com/wireapp/wire-server-deploy/pull/8/files#diff-e6b45cef94f317a441d6f606fa6e6d0c
@jschaul @fisx @neongreen @ChrisPenner @julialongtin

PR #11, opened 2019-02-07
Add some env config options

tiago-loureiro:
If `exim` and the STMP relay are not in the same network, you will end up seeing errors like _relay not permitted_; this allows us to tell exim which networks should be taken into account as relays

### DIFF
--- a/charts/demo-smtp/templates/deployment.yaml
+++ b/charts/demo-smtp/templates/deployment.yaml
@@ -23,8 +23,10 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           env:
-            - name: RELAY_NETWORKS
-              value: "{{ .Values.config.relayNetworks }}"
+        {{- range $key, $val := .Values.envVars }}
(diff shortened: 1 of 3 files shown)


charts/demo-smtp/templates/deployment.yaml
```diff
@@ -22,6 +22,9 @@ spec:
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          env:
+            - name: RELAY_NETWORKS
+              value: "{{ .Values.config.relayNetworks }}"
```
jschaul: Can you make this optional? (currently, if left unspecified, the chart does not deploy)

tiago-loureiro: Ah yes, meant to do so, thank u!

tiago-loureiro: I'd actually rather fix it like this https://github.com/wireapp/wire-server-deploy/pull/11/commits/9803a9f1a5af1af58e663526e058ecb2aa00a13f

PR #12, opened 2019-02-07
Add a docs/index.md file

neongreen:
Github Pages doesn't generate an index automatically (or it does but requires configuration; an explicit index might still be better)


neongreen:
It's possible to autogenerate a TOC (Jekyll supports a templating language with loops and whatever else), but if we want a nicer TOC than a list of links in random order, we'll spend more time writing Jekyll templates than anyone wants to.

PR #13, opened 2019-02-07
Working updated versions

tiago-loureiro:
~~Please don't merge this yet;~~ I have deployed this version of wire-server on our bare env and all looks good but I'd like to update the webapp versions too, which will require a few more tweaks on the charts.


tiago-loureiro:
This is now good to go

PR #21, opened 2019-02-13
Update sync.sh

tiago-loureiro:


bin/sync.sh
```diff
@@ -10,7 +10,7 @@
 # This script uses the helm s3 plugin,
 # for more info see https://github.com/hypnoglow/helm-s3
 
-# Usage: ./bin/sync.sh [--force]
+# Usage: ./bin/sync.sh [--force-push]
```
jschaul: The USAGE variable and comment a few lines down are a little duplicates, perhaps we can remove one of them?

tiago-loureiro: Ah yes; let's remove this whole comment then

PR #23, opened 2019-02-13
Allow defining TURN config as a list of URLs

tiago-loureiro:
This is pretty sufficient in many installations with static IP addresses; a more dynamic strategy can still be added later, as suggested [here](https://github.com/wireapp/wire-server-deploy/blob/master/README.md#status)

### DIFF
--- a/charts/brig/templates/deployment.yaml
+++ b/charts/brig/templates/deployment.yaml
@@ -31,33 +31,12 @@ spec:
         - name: "brig-config"
           configMap:
             name: "brig"
+        - name: "turn-servers"
+          configMap:
+            name: "turn"
         - name: "brig-secrets"
           secret:
             secretName: "brig"
-        - name: "turn-servers"
-          emptyDir: {}
-      initContainers:
-        # TODO: Proper TURN/calling config;
-        # this just creates a file to allow brig to start, but
-        # audio/video/TURN will not work
-        - name: fake-turn
-          image: alpine:3.6
-          command:
-          - sh
-          - -c
-          - echo -e "{{- range .Values.turnStatic.v1 }}{{ . }}\n{{- end }}" | sed '$d' > /etc/wire/brig/turn/turn-servers.txt
-          volumeMounts:
-          - name: turn-servers
-            mountPath: "/etc/wire/brig/turn"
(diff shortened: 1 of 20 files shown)


charts/brig/templates/deployment.yaml
```diff
@@ -45,7 +45,7 @@ spec:
           command:
           - sh
           - -c
-          - echo turn:127.0.0.1:3478 >> /etc/wire/brig/turn/turn-servers.txt
+          - echo -e "{{- range .Values.turnStatic.v1 }}{{ . }}\n{{- end }}" | sed '$d' > /etc/wire/brig/turn/turn-servers.txt
```
jschaul: instead of an empty dir "turn-servers" and an echo init container, wouldn't it be cleaner to create a second configmap and mount that?
(adapt the TODO comment above, and ensure to add another checksum/turn-config annotation)

charts/wire-server/requirements.yaml
```diff
@@ -58,7 +58,7 @@ dependencies:
     - haskellServices
     - services
 - name: brig
-  version: "0.1.2"
+  version: "0.1.3"
```
jschaul: let's perhaps keep all versions for all wire-server charts in sync, run `set-version.sh wire-server 0.1.3` - otherwise this change won't be published.

charts/brig/templates/deployment.yaml
tiago-loureiro: Yes, that's much better, will do so.

charts/brig/values.yaml
```diff
@@ -68,6 +68,13 @@ config:
   smtp:
     passwordFile: "/etc/wire/brig/secrets/smtp-password.txt"
 
+turnStatic:
```
jschaul:
```suggestion
# For audio/video calling, override this turnStatic config with DNS or IPs of your TURN server
# (the default here is necessary for the brig executable to start, but audio/video calling will not work with this default)
turnStatic:
```

tiago-loureiro:
> Could you just run `./bin/set-version.sh wire-server 0.1.3` manually before merging this PR (until versioning/syncing is fully automated)?

Yup, you are too fast in reviewing! ;)

PR #26, opened 2019-02-20
Minimal setup

tiago-loureiro:
In this PR:
 * Improve the demo-setup script slightly to allow skipping the creation of a load balancer on the cluster as well as to skip dependency updates (to speed up helm chart installations - use if you know what u are doing)
 * Allows overriding `resources` for even more minimal demo setups (e.g, there's an installation currently running where `memory requests` are all `"100Mi"`)


neongreen:
Can the branch be deleted?

tiago-loureiro:
Yup, somehow missed it, thanks

PR #42, opened 2019-04-16
Improve aws ingress

fisx:
Tasks:

- [x] add elb templates for team settings, account pages, ~~minio-ephemeral-s3~~
- [x] enable and test team settings (needs private repo access, leave disabled by default)
- [x] enable and test account pages (needs private repo access, leave disabled by default)
- [x] ~~find a better solution to locate the ephemeral `fake-aws-s3` service (may require upstream changes, not sure)~~ -> taken care of in #63


fisx:
### background for 7966bc5

before:

```
$ kf describe pod webapp-5dc6bf74b7-wvxf5
Name:               webapp-5dc6bf74b7-wvxf5
[...]
Containers:
  webapp:
    [...]
    Environment:
      NODE_PORT:                            8080
      APP_BASE:                             https:///
      BACKEND_REST:                         https://kube-fisx-nginz-https.zinfra.io
      BACKEND_WS:                           wss://kube-fisx-nginz-ssl.zinfra.io
      APP_BASE:                             https://kube-fisx-webapp-https.zinfra.io/
      CSP_EXTRA_CONNECT_SRC:                https://*.zinfra.io, wss://*.zinfra.io
      [...]
```

after:

```
$ kf describe pod webapp-65948c6995-7rd26
Name:               webapp-65948c6995-7rd26
[...]
Containers:
  webapp:
    [...]
    Environment:
      NODE_PORT:                            8080
      APP_BASE:                             https://kube-fisx-webapp-https.zinfra.io/
      APP_BASE:                             https://kube-fisx-webapp-https.zinfra.io/
      BACKEND_REST:                         https://kube-fisx-nginz-https.zinfra.io
      BACKEND_WS:                           wss://kube-fisx-nginz-ssl.zinfra.io
      CSP_EXTRA_CONNECT_SRC:                https://*.zinfra.io, wss://*.zinfra.io
      [...]
```

i'm not sure it's good that we have both the helm value and the env var, but at least they are the same now.

i'm also not sure how the default value `staging-nginz.zinfra.io` could creep in when the *first* `APP_BASE` was empty, but the *second* was still good.  but it did.  now it's working.

tiago-loureiro:
There are some env vars already defined in the deployment, namely: https://github.com/wireapp/wire-server-deploy/blob/master/charts/webapp/templates/deployment.yaml#L33-L40

These are then merged in with the `envVars` defined in the values file. The idea behind this approach (allowing free form env vars) was to avoid having to deploy a new chart every time there is a change in the app.

Another point of this approach was that some env vars should be required while others optional. Perhaps we should indicate that in the values file.

> i'm also not sure how the default value staging-nginz.zinfra.io could creep in when the first APP_BASE was empty, but the second was still good. but it did. now it's working.

Right, so it's defined twice: one by the deployment, one by the defined envVars.

tiago-loureiro:
I have rebased fixed this and tried it against our kube-roles env, works!

charts/aws-ingress/values.yaml
```diff
@@ -44,3 +45,21 @@ ingress:
...
+      hostname: <prefix>-teams-https.<domain>
+      ttl: 300
+    http:
+      teamSettingsPort: 8080
+  accountPages:
+    https:
+      externalPort: 443
+      sslCert: arn:aws:iam::00000-accountnumber-00000:server-certificate/example.com
+      sslPolicy: ELBSecurityPolicy-TLS-1-2-2017-01
+      hostname: <prefix>-account-https.<domain>
```
jschaul: Let's change these hostnames to be consistent with documentation: https://docs.wire.com/how-to/install/helm.html#how-to-set-up-dns-records.
I.e. use `example.com` instead of `<domain>` for easier grepping/chaging, and remove the `<prefix>`

(Also please apply the above to all items in aws-ingress/values.yaml)

charts/aws-ingress/values.yaml
```diff
@@ -44,3 +45,21 @@ ingress:
     selector:
       key: app
       value: minio # (currently) fake-aws-s3 chart uses 'minio', minio-external chart uses 'minio-external'
+  teamSettings:
+    https:
+      externalPort: 443
+      sslCert: arn:aws:iam::00000-accountnumber-00000:server-certificate/example.com
+      sslPolicy: ELBSecurityPolicy-TLS-1-2-2017-01
```
jschaul: The security policy isn't as good as it can be, let's use a custom one that is a bit more secure as on our own production, please.

charts/aws-ingress/values.yaml
tiago-loureiro: Good point, I changed a couple more references: https://github.com/wireapp/wire-server-deploy/commit/e1ebe2d429deed6172e09806859ce204db4d64e1

charts/aws-ingress/values.yaml
tiago-loureiro: I don't think this is possible using the k8s API though, at least not when looking at: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go

I.e., I don't see a way to specify custom protocols (i.e., custom policy attributes). Happy to be proved wrong 😬

jschaul: You're right that custom policy attributes seem no longer supported by AWS on the new types of load balancers at all anymore. You could use `ELBSecurityPolicy-FS-1-2-Res-2019-08` though, which is the best AWS currently offers. Search for `ELBSecurityPolicy-FS` on [this page](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies)

tiago-loureiro:
As discussed, and since this creates Classic ELB's we'll merge this as-is due to the fact that the suggested security policies are not supported on the classic ELB's

PR #55, opened 2019-05-13
Basic backoffice tool

tiago-loureiro:
The backoffice chart provides the [following functionality](https://github.com/wireapp/wire-server-deploy/blob/backoffice/charts/backoffice/README.md).

Related PR https://github.com/wireapp/wire-server/pull/757


charts/backoffice/README.md
```diff
@@ -0,0 +1,11 @@
+Backoffice frontend
+===================
+
+This chart provides a basic frontend app that is composed of nginx serving swagger and will soon be found here [here](https://github.com/wireapp/wire-server/blob/develop/tools/backoffice-frontend/README.md). It serves as a tool to perform operations on users and teams such as visualising their user profiles, suspending or even deleting accounts. It is used internally at Wire to provide customer support the means to respond to certain queries from our customers and can be used by anyone that decides to deploy it on their cluster(s).
+
+It is intended to be accessed, at the moment, only by means of port forwarding and therefore only available to cluster admins (or more generally, clusters users able to port forward).
```
jschaul:
```suggestion
It is intended to be accessed, at the moment, only by means of port forwarding and therefore only available to cluster admins (or more generally, clusters users able to port forward).

:warning: **DO NOT expose this chart to the public internet** with an ingress - doing so would give anyone the ability to read part of the user database, delete users, etc.
```

charts/backoffice/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Once the chart is installed, and given default values, you can access the frontend with 2 steps:
+
+ * kubectl port-forward svc/backoffice 8080:8080
+ * Open your local browser at http://localhost:8080
```
jschaul: It might be useful to create a [notes.txt](https://github.com/helm/helm/blob/master/docs/chart_template_guide/notes_files.md) file.

charts/backoffice/values.yaml
```diff
@@ -0,0 +1,29 @@
...
+  internalPort: 8080
+  externalPort: 8080
+# The `baseUrl` is the URL that you plan to use to access
+# the Backoffice tool, i.e., the address that you plan to
+# use to access the tool from. E.g., if you want to access
+# it from localhost:8080, then you should define baseUrl
+# as `http://localhost:8080` and port-forward the service
+# from port 8080 to the defined externalPort. E.g.,
+# kubectl port-forward svc/backoffice 8080:8080
+# You can then access the tool locally at http://localhost:8080
```
jschaul: Comment rewrite suggestion:

```yaml
# `baseUrl` is the URL by which the backoffice is accessed in the browser.
# It needs to match exactly for links to work correctly (it's used in the swagger UI).
# Example: if you want to access it from localhost:8080, then you should define baseUrl
# as `http://localhost:8080` and port-forward the service from port 8080 to the defined externalPort:
#   kubectl port-forward svc/backoffice 8080:8080
```

charts/backoffice/README.md
tiago-loureiro: 👍

tiago-loureiro:
> Why helm install stern; helm install backoffice if one command would do?

Even with 2 charts, we install the tool by simply running `helm install backoffice`

> There's currently no use case to keep them separate, or is there?

No, u are right - deploying `stern` on its own is not very interesting though, can totally make this a single chart, let me give that a go.

tiago-loureiro:
_11 changed file with 167 additions and 297 deletions._

Making it all a single chart sounds much better 👌

charts/backoffice/templates/NOTES.txt
```diff
@@ -1,4 +1,6 @@
 You have successfully deployed the backoffice tool! Given the default config, you can now access the frontend in 2 steps:
 
-  * kubectl port-forward svc/backoffice 8080:8080
+  * kubectl --namespace {{ .Release.Namespace }} port-forward svc/backoffice 8080:8080
```
tiago-loureiro: Ah, I didn't know u could do this on `NOTES.txt`, thanks 👍

PR #63, opened 2019-05-21
add s3-minio AWS ELB support

jschaul:
Can be useful to test minio on AWS; or, for ephemeral environments, to avoid configuring secrets and buckets on real S3.
(disabled by default - when using AWS, one has access to real S3)

### DIFF
--- a/charts/aws-ingress/values.yaml
+++ b/charts/aws-ingress/values.yaml
@@ -43,4 +43,4 @@ ingress:
       s3Port: 9000
     selector:
       key: app
-      value: minio
+      value: minio # (currently) fake-aws-s3 chart uses 'minio', minio-external chart uses 'minio-external'
--- a/charts/aws-ingress/values.yaml
+++ b/charts/aws-ingress/values.yaml
@@ -31,7 +31,7 @@ ingress:
       ttl: 300
     ws:
       wsPort: 8081
-  s3-minio:
+  s3minio:
     enabled: false # set to true if you wish to use minio on AWS instead of using real S3
     https:
       externalPort: 443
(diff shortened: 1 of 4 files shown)


jschaul:
(re-uses part of https://github.com/wireapp/wire-server-deploy/pull/42)

charts/aws-ingress/values.yaml
```diff
@@ -31,3 +31,16 @@ ingress:
...
+    enabled: false # set to true if you wish to use minio on AWS instead of using real S3
+    https:
+      externalPort: 443
+      sslCert: arn:aws:iam::00000-accountnumber-00000:server-certificate/example.com
+      sslPolicy: ELBSecurityPolicy-TLS-1-2-2017-01
+      hostname: <prefix>-s3-https.<domain>
+      ttl: 300
+    http:
+      s3Port: 9000
+    selector:
```
tiago-loureiro: Is there any reason to make this configurable vs. hardcoding it in the service?

jschaul: Currently, fake-aws-s3 uses `app: minio`, and minio-external uses `app: minio-external`. We don't use `app` anywhere else though (and use the `wireService` selector), so I figured some flexibility here would be good. Do you think it's worth removing the selector key from values (hardcoding to `app`) ?

tiago-loureiro: >  fake-aws-s3 uses app: minio, and minio-external uses app: minio-external

OK, yes that's what I thought the reasoning was.

>  Do you think it's worth removing the selector key from values (hardcoding to app) ?

Nope, looks good, was wondering the reasoning behind it!

PR #64, opened 2019-05-21
Update chart version for minio dependency

jschaul:
untested if everything still works. May fix https://github.com/wireapp/wire-server-deploy/pull/46 (which has no effect currently).


jschaul:
CI tests passed so the chart upgrade should not have broken anything.

PR #69, opened 2019-06-04
Feature/ansible configuration cassandra

jschaul:
playbook & role to install cassandra, including the ability to use a different network interface for binding and for SSH access.


jschaul:
Depends on https://github.com/wireapp/ansible-cassandra/pull/8 being merged and tagged with `v0.1.2`.

PR #71, opened 2019-06-04
Feature/ansible configuration binaries

jschaul:
A way to download `helm` and `kubectl` cli binaries of the correct version.


julialongtin:
Worked right the first time. Nice work!

PR #76, opened 2019-06-11
redis support, i think?

julialongtin:


arianvp:
We use the `redis` chart for quite a while now.  If we need persistence we'll use  kubernetes' PersistentVolume support

PR #77, opened 2019-06-11
Feature/ansible configuration restund

tiago-loureiro:
Basic playbook to install restund servers


tiago-loureiro:
> I made a few tweaks, I hope you don't mind. Works!

Nope, looks great; thanks!

PR #78, opened 2019-06-11
[WIP] Initial

tiago-loureiro:
Still requires a bit more testing, particularly with:
```
minio_network_interface: "{% if minio_network_interface is not defined %}{% if ansible_default_ipv4 is defined %}{{ ansible_default_ipv4.interface }}{% else %}eth0{% endif %}{% else %}{{ minio_network_interface }}{% endif %}"
```
but should not be too far from this.


jschaul:
superseded by #80

PR #80, opened 2019-06-12
Minio

julialongtin:


ansible/README.md
```diff
@@ -141,6 +142,22 @@ elasticsearch_network_interface=ens3
 poetry run ansible-playbook -i hosts.ini elasticsearch.yml -vv
 ```
 
+### Minio
+
+* In your 'hosts.ini' file, in the `[minio:vars]` section, set 'minio_network_interface'. For example:
```
jschaul: This should state that specifying a network interface is mandatory (the defaults don't work correctly).

ansible/minio.yml
```diff
@@ -0,0 +1,45 @@
...
+
+# Commented out for now. should not depend on hostname.
+#    - role: hostname
+#      tags:
+#        - hostname
+
+# Also commented out. ntp broken?
+#     - role: ansible-role-ntp
+#       tags:
+#         - ntp
```
jschaul: Feel free to remove these comments entirely.

tiago-loureiro:
Any reason why we don't use `hostname` and `ntp` roles here?

jschaul:
> Any reason why we don't use hostname and ntp roles here?

Do we need those? I guess it's orthogonal to minio. Does minio need ntp?

PR #84, opened 2019-06-12
A more production like setup

tiago-loureiro:
- [ ] Better documentation
- [x] Add minio external after testing


tiago-loureiro:
Should be ready to be reviewed; tried to add `CHANGEME-PROD` in all places where it requires attention, feel free to double check if we missed sth.

PR #91, opened 2019-06-21
Fix: url trailing slashes

fisx:
Without this fix, `/join/?key=....` urls have two slashes in front of `/join`, which causes a 404.

Related: https://github.com/wireapp/wire-server-deploy/blob/master/charts/webapp/templates/deployment.yaml#L35-L36

It's a bit unfortunate that APP_BASE can't be overwritten, and that `//` handling is so brittle.  Not sure how to fix this, though.  Perhaps remove these `env` entries entirely where they can be declared explicitly, rather than fixing them a little for this instance?

### DIFF
--- a/charts/webapp/templates/deployment.yaml
+++ b/charts/webapp/templates/deployment.yaml
@@ -34,11 +34,11 @@ spec:
           - name: NODE_PORT
             value: "{{ .Values.service.http.internalPort }}"
           - name: APP_BASE
-            value: https://{{ .Values.config.externalUrls.appHost }}
+            value: "https://{{ .Values.config.externalUrls.appHost }}"
           - name: BACKEND_REST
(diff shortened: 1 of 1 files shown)


tiago-loureiro:
> It's a bit unfortunate that APP_BASE can't be overwritten, and that // handling is so brittle. Not sure how to fix this, though. Perhaps remove these env entries entirely where they can be declared explicitly, rather than fixing them a little for this instance?

We can specify them differently, yes; let's fix this across the board and let's figure out what will happen to existing installations.

What is important here is that defining `NODE_PORT`, `APP_BASE`, `BACKEND_{REST|WS}` is the bare minimum to have a working app so we should just fail if they don't get defined.

jschaul:
While this can further be cleaned up, this change of this PR is correct as it is: webapp requires no trailing slash, team/account pages still do require a trailing slash.

PR #93, opened 2019-06-21
nginx-ingress increased buffer size

jschaul:
502 due to 'upstream sent too big header while reading response header
from upstream'

See https://andrewlock.net/fixing-nginx-upstream-sent-too-big-header-error-when-running-an-ingress-controller-in-kubernetes/


jschaul:
(the proxy-buffer-size is the relevant (and important) line)

PR #96, opened 2019-06-25
Ansible restund example root and tls and download if needed

tiago-loureiro:
Uses the latest available version from the ansible restund playbook.

### DIFF
--- a/ansible/restund.yml
+++ b/ansible/restund.yml
@@ -1,4 +1,22 @@
 ---
+- name: gather tls certificate
+  hosts: localhost
+  pre_tasks:
+    - name: Read TLS config
+      set_fact:
+        # Reminder that the pem file should look like:
+        # -----BEGIN CERTIFICATE-----
+        # --- ... CERT CONTENT ... --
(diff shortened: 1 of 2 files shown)


ansible/restund.yml
```diff
@@ -1,4 +1,22 @@
...
+        # --- ... CERT CONTENT ... --
+        # -----END CERTIFICATE-----
+        # -----BEGIN CERTIFICATE-----
+        # --- ... INTERMEDIATE ..----
+        # -----END CERTIFICATE----
+        # -----BEGIN PRIVATE KEY-----
+        # --- .... PRIV KEY -----
+        # -----END PRIVATE KEY-----
+
+        restund_tls_certificate: "{{ lookup('file', '/tmp/tls_cert_and_priv_key.pem') }}"
```
jschaul: You can move this line to the `hosts: restund` block and delete the localhost block entirely. `lookup` plugins are executed on localhost by default.

tiago-loureiro: Oh, OK. Thanks!

PR #99, opened 2019-06-25
Feature/docker registry

jschaul:


tiago-loureiro:
Think ansible/files/registry/images.sh can be a bit simplified, we can do that later 👌

PR #105, opened 2019-07-02
[MAYBE-DONT-MERGE] mitogen

jschaul:
This uses mitogen for (possible) ansible speedups - the actual speedup for the playbooks we have has not been measured yet (TODO) - if it's insignificant, it might not be worthwhile, as it seems a little brittle (not working with ansible 2.8, requiring a specific path to the mitogen strategy directory, etc).


jschaul:
outdated.

PR #106, opened 2019-07-04
Adjustments i had to make to make things work

arianvp:


ansible/iptables.yml
```diff
@@ -0,0 +1,27 @@
+- hosts: kube-node
+  tasks:
+    - name: "Redirect TLS traffic to the kubernetes ingress"
+      become: true
+      iptables:
+        table: "nat"
+        chain: "PREROUTING"
+        in_interface: "eth0"
```
tiago-loureiro: This will not work on instances where the interface is `eth0`; can be improved later, probably taking the interface name from the hosts file?

shell.nix
```diff
@@ -0,0 +1,8 @@
+{ pkgs ? import <nixpkgs> {}}:
```
jschaul: Perhaps we can place this on gitignore?

PR #112, opened 2019-07-16
improve helm chart sync.sh script

jschaul:
* Remove workaround part of sync.sh script (removes the need to copy files from one S3 dir to another)
* version lock helm s3 plugin (to a version of a fork of the original author)

### DIFF
--- a/bin/sync.sh
+++ b/bin/sync.sh
@@ -14,7 +14,7 @@ set -eo pipefail
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-USAGE="Sync helm charts to S3. Usage: $0 to sync all charts or $0 <chartname> to sync only a single one. --force-push can be used to override S3 artifacts."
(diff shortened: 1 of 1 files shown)


bin/sync.sh
```diff
@@ -28,49 +28,24 @@ if [ -n "$chart_name" ] && [ -d "$SCRIPT_DIR/../charts/$chart_name" ]; then
...
-    done
-
-    # update and upload index to charts
-    sed -i'.bak' -e 's/s3\:\/\//https\:\/\/s3-eu-west-1.amazonaws.com\//g' index.yaml
-    aws s3 cp index.yaml s3://public.wire.com/$PUBLIC_DIR/index.yaml
-}
+# See https://github.com/hypnoglow/helm-s3/pull/56 for reason to use fork
+s3_plugin_version=$(helm plugin list | grep "^s3 " | awk '{print $2}')
+if [[ $s3_plugin_version != "0.9.0" ]]; then
+    echo "not version 0.9.0 from steven-sheehy fork, upgrading or installing plugin..."
```
julialongtin: isn't this a little fragile? what happens if upstream releases a 0.9.0 that does not have this feature?

jschaul: Fragility fixed in last commit, where I also check the git hash.

jschaul:
Also added `--reindex` option.

PR #116, opened 2019-07-26
Update and comment token timeouts

tiago-loureiro:
1 day is a better default for `sessionTokenTimeout` (also in line with what we are using ourselves).
In addition, added some comments to help out understand what the numbers mean


jschaul:
> added some comments to help out understand what the numbers mean

(Unfortunately, usage of `yq` strips all comments - and CI is updating this file)

PR #128, opened 2019-08-16
script for updating docker images in offline environment

jschaul:
Adds a script for performing updates in an offline environment:

The download script downloads docker images to a directory, and adds a bash script which can be used, in an offline environment, to load those images to a local docker registry.


jschaul:
ready for review

bin/offline/online-download-images.sh
```diff
@@ -0,0 +1,76 @@
...
+mkdir "$FOLDER"
+
+# now that versions are known, write load_into_registry script
+echo "#!/usr/bin/env bash
+
+set -ex
+
+for image in *.tar; do
+    docker load < \"\$image\"
+done
```
tiago-loureiro: What are these tarballs?

jschaul: There are two scripts (the main script writes the second script):

the main script does `docker save quay.io/wire/brig:2.60.0 > brig.tar`

The second script (created by the main one) then does `docker load < brig.tar`

Does this need more comments to be clear?

jschaul: added comments in https://github.com/wireapp/wire-server-deploy/pull/128/commits/0825a1ffd8e096f34b57729df19816bffdb69177

PR #132, opened 2019-08-20
Remove AWS specifics from wire-server-metrics

tiago-loureiro:
The given default values could not be overwritten using `helm` (they can only be removed), forcing users of this chart to use AWS (or using ephemeral).

This diff removes nearly all the defaults and turns those into documentation; if this looks good, I will propose chances in the docs [here](https://github.com/wireapp/wire-server-deploy/blob/master/docs/monitoring.md) too.


tiago-loureiro:
I believe that I wrongly interpreted this chart partially due to a too old `helm` version and slightly incorrect docs - another PR coming up instead

PR #134, opened 2019-08-22
etcd administrating helper scripts, more bash scripts

jschaul:
* adds some helper bash scripts (for use with CI or in offline environments)
* adds etcd helper scripts as part of the installation of kubernetes:

```
root@node1:~# etcd-health.sh 
Cluster-Endpoints: https://127.0.0.1:2379
cURL Command: curl -X GET https://127.0.0.1:2379/v2/members
member 7c37f7dc10558fae is healthy: got healthy result from https://10.10.1.11:2379
member cca4e6f315097b3b is healthy: got healthy result from https://10.10.1.10:2379
member e767162297c84b1e is healthy: got healthy result from https://10.10.1.12:2379
cluster is healthy
```
   
The following is annoying:

```
root@node1:~# etcdctl member list
client: etcd cluster is unavailable or misconfigured; error #0: dial tcp 127.0.0.1:4001: getsockopt: connection refused
; error #1: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
```

So the helper script sets the right parameters for convenience:

```
root@node1:~# etcdctl3.sh member list
7c37f7dc10558fae, started, etcd_1, https://10.10.1.11:2380, https://10.10.1.11:2379
cca4e6f315097b3b, started, etcd_0, https://10.10.1.10:2380, https://10.10.1.10:2379
e767162297c84b1e, started, etcd_2, https://10.10.1.12:2380, https://10.10.1.12:2379
```


jschaul:
ready for review

bin/set-wire-server-image-version.sh
```diff
@@ -0,0 +1,13 @@
...
+target_version=${1?$USAGE}
+
+charts=(brig cannon galley gundeck spar cargohold proxy cassandra-migrations)
+
+for chart in "${charts[@]}"; do
+    sed -i "s/  tag: .*/  tag: $target_version/g" "charts/$chart/values.yaml"
+done
+
+#special case nginz
+sed -i "s/   tag: .*/   tag: $target_version/g" "charts/nginz/values.yaml"
```
tiago-loureiro: What's different here, number of spaces, right? Would it make sense to use a tool like `yaml`? Or are we trying to avoid more dependencies?

ansible/roles/etcd-helpers/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+# TODO - add this task/role to kubespray upstream, potentially.
```
tiago-loureiro: That sounds cool, let's do it :)

ansible/kubernetes.yml
```diff
@@ -1,4 +1,8 @@
 ---
 # This assumes you ran "make download" to download the kubespray files
-# 
+#
 - import_playbook: roles-external/kubespray/cluster.yml
+
+- hosts: etcd
```
tiago-loureiro: By default, `etcd` is not part of the hosts, right?

bin/set-wire-server-image-version.sh
jschaul: Just number of spaces.

We're trying to avoid more dependencies - this script was originally written on a system without `yaml` available (but `sed` was available).

ansible/kubernetes.yml
jschaul: https://github.com/wireapp/wire-server-deploy/blob/master/ansible/hosts.example.ini#L83

it is part by default, see above. Kubespray needs this to be in the hosts file.

tiago-loureiro: 🤦‍♂ I looked in the file and missed it. Move along. Thanks!

PR #139, opened 2019-08-27
legal hold tokens

jschaul:
* for integration tests, brig now also depends on nginz
* nginz acl file updated
* whitelisting /access for user tokens has no effect, as that route is not zauth-authenticated

Needed for https://github.com/wireapp/wire-server/pull/761

### DIFF
--- a/charts/nginz/conf/static/zauth.acl
+++ b/charts/nginz/conf/static/zauth.acl
@@ -4,8 +4,6 @@ a (blacklist (path "/provider")
              (path "/bot/**")
              (path "/i/**"))
 
-u (whitelist (path "/access"))
-
 b (whitelist (path "/bot")
              (path "/bot/**"))
 
--- a/charts/nginz/conf/static/zauth.acl
+++ b/charts/nginz/conf/static/zauth.acl
@@ -12,3 +12,7 @@ b (whitelist (path "/bot")
 p (whitelist (path "/provider")
              (path "/provider/**"))
 
+# LegalHold Access Tokens
+la (whitelist (path "/notifications")
+              (path "/assets/v3/**")
+              (path "/users/**"))
--- a/charts/brig/templates/tests/configmap.yaml
+++ b/charts/brig/templates/tests/configmap.yaml
@@ -20,6 +20,10 @@ data:
       host: cargohold
       port: 8080
 
+    nginz:
+      host: nginz
+      port: 8080
+
     provider:
       privateKey: /etc/wire/integration-secrets/provider-privatekey.pem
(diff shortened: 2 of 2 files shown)


charts/nginz/conf/static/zauth.acl
```diff
@@ -4,11 +4,13 @@ a (blacklist (path "/provider")
              (path "/bot/**")
              (path "/i/**"))
 
-u (whitelist (path "/access"))
```
jschaul: whitelisting /access for user tokens has no effect, as that route is not zauth-authenticated

tiago-loureiro: > whitelisting /access for user tokens has no effect, as that route is not zauth-authenticated

OK, I was wondering about this - Does this not mean though that this is the only route where user tokens are allowed/accepted?

jschaul: As far as I understand:

* User Tokens are always sent inside a cookie header. Access Tokens are always sent either as an `Authorization: Bearer <token>` or a query paramater.
* nginz does nothing with Cookies. Cookies are just passed through and remain untouched by nginz's zauth module.
* It just happens that the cookies generated in brig contain `path=/access`, meaning the browser only sends them to /access, not to other routes. 
* I don't know what mobile clients do. Perhaps they send cookies elsewhere, perhaps not, we don't know as we don't check for them.

For stuff in the `Authorization: ` header, nginz zauth looks at them: https://github.com/wireapp/wire-server/blob/develop/services/nginz/third_party/nginx-zauth-module/zauth_module.c#L303-L323 (if it doesn't find any authroization header it also looks for a query parameter, and if that is also not there it apparently looks for a `zprovider` cookie. But it doesn't look like the line I removed would have any effect. I admit I only have a partial understanding of zauth, so I might be wrong

charts/nginz/conf/static/zauth.acl
```diff
@@ -4,11 +4,13 @@ a (blacklist (path "/provider")
...
-
 b (whitelist (path "/bot")
              (path "/bot/**"))
 
 p (whitelist (path "/provider")
              (path "/provider/**"))
 
+# LegalHold Access Tokens
+la (whitelist (path "/notifications")
+              (path "/assets/v3/**")
```
arianvp: Why does legalhold need access to the assets API?

arianvp: Oh I guess it makes sense to have access to it

jschaul: (to get not just text messages, but also e.g. pictures)

PR #140, opened 2019-09-03
Make the update.sh script take a _path_ instead of a _name_

arianvp:
This means I can do:

`/bin/update.sh charts/wire-se<TAB>`  and it auto-completes to the right
chart name, which saves me 10 seconds per day and makes me not have to
know a chart name apriori. I think the UX is a lot better, as we get
free completion of what charts are available

Not a really important change, just something that made things a bit easier for me when working with these scripts interactively

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ If submitting pull requests, please follow these guidelines:
 
 * if you want to make larger changes, it might be best to first open an issue to discuss the change.
 * if helm charts are involved, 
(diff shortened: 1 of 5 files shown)


arianvp:
Are there any other scripts we use internally that depend on this script that might break because of this?

jschaul:
> Are there any other scripts we use internally that depend on this script that might break because of this?

I don't think so, I did a quick grep, didn't find anything.

But actually I noticed in this repo `bin/sync.sh` uses it, so the path there should be adjusted before merging.

arianvp:
@jschaul oould we look at that script together tomorrow?  It's a bit daunting to me :)

jschaul:
@arianvp please merge :)

PR #144, opened 2019-09-11
See https://github.com/wireapp/wire-server/pull/850

fisx:


charts/galley/values.yaml
```diff
@@ -23,5 +23,17 @@ config:
     maxTeamSize: 500
     maxConvSize: 500
     featureFlags:
-      sso: false
-      legalhold: false
\ No newline at end of file
+      # SSO: this sets the default setting for each time, which can
+      # always be overridden by customer support / backoffice.
+      # IMPORTANT: if you change sso from 'enabled' to 'disabled' after
+      # running 'enabled' in production, you need to run this migration
```
jschaul: same here. does `enabled` mean `enabled-by-default` ?

charts/galley/values.yaml
```diff
@@ -23,5 +23,17 @@ config:
...
+      # script to fix all teams that have registered an idp:
+      # https://github.com/wireapp/wire-server/tree/master/tools/db/migrate-sso-feature-flag
+      # if you don't, the idp will keep working, but the admin won't be
+      # able to register new idps.
+      sso: disabled-by-default
+      # Legal Hold: this decides whether customer support / backoffice
```
jschaul: This comment also doesn't seem fully correct anymore.

Can you be explicit about what the possible strings are that this feature can be set to? To this and also the `sso` flag above?

PR #146, opened 2019-09-16
Reference docs for feature flags.

fisx:


fisx:
weird.  i thought i fixed that in https://github.com/wireapp/wire-server/pull/848?

jschaul:
(Merge conflicts appeared due to CI using https://github.com/kislyuk/yq which doesn't preserve comments in yaml files. We now switched CI to using sed, which doesn't have that problem (it has other problems)).

PR #155, opened 2019-09-24
Feature/deeplinks

jschaul:
Adds an ansible role creating files for deeplinking in custom envrironments. Also adds version blacklisting files.


ansible/roles/minio-static-files/defaults/main.yml
```diff
@@ -0,0 +1,24 @@
...
+
+# FUTUREWORK:
+# Note the website, for Wire, points to https://wire.com
+# There is currently no equivalent for custom backends
+# So all URLs in the mobile app that point to a url on the wire website will
+# not work, as the default of 'www' points nowhere.
+websiteURL: "https://{{ prefix }}www.{{ domain }}"
+
+# This currently assumes android clients only
+blackListURL: "{{ assetsURL }}/public/blacklist/android.json"
```
tiago-loureiro: Why not just use `{{ assetsURL }}/public/blacklist` ?

jschaul: I briefly looked into the code of android, and they seem to have `"blacklistHost": "https://clientblacklist.wire.com/prod/android"`, i.e. the full path. If that changes, we can adapt this to be only a prefix.

PR #156, opened 2019-09-25
Minor updates

julialongtin:


ansible/Makefile
```diff
@@ -9,7 +9,7 @@ download-kubespray:
 	poetry run ansible-playbook -i localhost, -c local download_kubespray.yml
 
 .PHONY: download-cli-binaries
-download-cli-binaries:
+download-cli-binaries: download-ansible-roles
```
jschaul: if you add that dependency, I suppose you can remove the `download-ansible-roles` dependency from the `download` task above, to avoid having it download things twice.

julialongtin:
Make is smart enough not to rebuild a phony target. try this:
```
julial@bk-sandbox:~/kvm2$ cat Makefile
all: this that theother

this: theother
        echo this

that: theother
        echo that

theother:
        echo theother
julial@bk-sandbox:~/kvm2$ make
echo theother
theother
echo this
this
echo that
that
```

jschaul:
Oh, I see. I underestimated Make's smartness. Thanks!

PR #160, opened 2019-10-01
avs-nwtest: implement via jsonnet/kubecfg (WIP)

q3k:
This implements the deployment of the AVS Network Test Tool (avs-nwtesttool).

To make it a bit more interesting, we deploy it using kubecfg/jsonnet instead of using Helm.

For that, we pull in `kube.libsonnet` from bitnami, and write a small framework that is composed of:
 - wire deployments/environments
   (we call them deployment at the top-level configs, but environments deepr within jsonnet machinery to not confuse them with Kubernetes Deployments)
 - wire components
   (eg. microservices composed usually of a deployment+service that is exposed via nginz - this is a paradigm that can be extended in the future)

We not only export kubernetes-consumable objects from the jsonnet config, but also nginz routes as an example of multiple outputs from a single configuration.

This is all currently documented in a small README.md, which will be turned into real documentation once this PR stabilizes.

Fixes https://github.com/zinfra/backend-issues/issues/892

### DIFF
--- a/kube/lib/wire.libsonnet
+++ b/kube/lib/wire.libsonnet
@@ -1,0 +1,181 @@
+local kube = import "kube.libsonnet";
+
+// Wire deployment.
+
+{
+    local wire = self,
+
+    // A Wire Environment. Also called a Deployment.
+    // Contains all services/components required to run Wire on k8s, apart from
+    // per-cluster services.
+    // All components will be run in a given k8s namespace, by default named
+    // after the environment.
+    Environment(name):: {
+        local env = self,
+        local cfg = env.cfg,
+
+        // User configurable.
+        cfg:: {
+            // Name of environment, used internally. Can be overridden.
+            name: name,
+            // Kubernetes namespace in which environment will be run. Can be overridden.
+            namespace: name,
+
+            // Public DNS names to be used by wire-server.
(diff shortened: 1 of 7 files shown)


q3k:
For those unfamiliar with jsonnet: There is a tutorial available on their homepage, at https://jsonnet.org/ .

kube/lib/wire.libsonnet
```diff
@@ -0,0 +1,181 @@
...
+
+            // Public DNS names to be used by wire-server.
+            dns: {
+                // Top-level domain. Must be set, or all specific domain names
+                // underneath must be overriden.
+                domain: error "dns.domain must be set", // .my-wire.example.com
+
+                // Default DNS prefixes for top-level domain.
+                nginzHTTPS: "nginz-https.%s" % [cfg.dns.domain],
+                nginzTLS: "nginz-tls.%s" % [cfg.dns.domain],
```
jschaul: You decided to diverge from `nginz-ssl` used in other places :) 

I suppose we should find a naming consensus that we're all happy with.

jschaul:
This particular PR is obsolete. Migrating away from using helm to kubecfg or otherwise is a separate discussion.

PR #161, opened 2019-10-07
*: unify DNS names for demo deployments

q3k:
This change unifies the list of domains that users should set up to the
following list of canonical prefixes:

 - `nginz-https.<domain>`
 - `nginx-ssl.<domain>`
 - `webapp.<domain>`
 - `s3.<domain>`
 - `teams.<domain>` (optional)
 - `account.<domain>` (optional)

Naturally, following this scheme is not a hard requirement, and users can
change these as they please. However, keeping them in sync allows for
users to do a blind search/replace from `example.com` to their domain,
without having to immediately understand the roles of different
subdomains.

Optimally we would have all of these defined in a single source of truth
file that would be sourced from all Helm charts, but I'm not sure this
is something that we can easily do.

### DIFF
--- a/ansible/roles/minio-static-files/defaults/main.yml
+++ b/ansible/roles/minio-static-files/defaults/main.yml
@@ -6,7 +6,7 @@
 #domain: example.com
 #deeplink_title: Example Environment
 
-assetsURL: "https://{{ prefix }}s3.{{ domain }}"
+assetsURL: "https://{{ prefix }}assets.{{ domain }}"
 deeplink_config_json: "{{ assetsURL }}/public/deeplink.json"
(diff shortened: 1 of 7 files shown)


jschaul:
We could, additionally, settle on `assets` as the default name, rather than `s3`. Assets is the term used in [our code](https://github.com/wireapp/wire-server/blob/68768fc7d1cb05b50eae4895605cef2e1adfcac1/services/cargohold/src/CargoHold/API.hs#L55), and is less of an amazon-web-services brand name.

q3k:
Good point! I have renamed `s3` to `assets`. PTAL.

PR #171, opened 2020-01-07
Separate etcd from k8s nodes & moves them into dedicated machines

lucendio:
* before, control-plane + etcd + workload were sitting in the same machines,
  but this does not reflect the current architectural design as described
  in the docs (e.g. https://docs.wire.com/how-to/install/planning.html#implementation-planl)
  This change aims to apply the described design to TF example
* fixes some typos
* states a note on why kubespray is not installed as a role through
  requirements.yml

### DIFF
--- a/ansible/download_kubespray.yml
+++ b/ansible/download_kubespray.yml
@@ -1,4 +1,6 @@
 # download a specific version of kubespray
+# Note: installing it via ansible-galaxy would re-arranges its content (roles/etcd is misplaced)
+#       the reason for this behaviour is yet to be known
 # Usage: see Makefile
 - name: download kubespray
   hosts: localhost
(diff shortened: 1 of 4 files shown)


ansible/download_kubespray.yml
```diff
@@ -1,4 +1,6 @@
 # download a specific version of kubespray
+# Note: installing it via ansible-galaxy would re-arranges its content (roles/etcd is misplaced)
```
jschaul:
```suggestion
# Note: installing it via ansible-galaxy would re-arrange its content (roles/etcd is misplaced)
```

lucendio:
fixed the spelling and applied same change to `ansible/hosts.example.ini` too, since I just stumbled upon it while looking at `hosts.example-demo.ini`. @jschaul, mind giving it another quick look?

PR #185, opened 2020-02-11
New option setUserMaxPermClients is now available for brig

tiago-loureiro:
Folow up to https://github.com/wireapp/wire-server/pull/969

### DIFF
--- a/charts/brig/values.yaml
+++ b/charts/brig/values.yaml
@@ -69,7 +69,7 @@ config:
     setDeleteThrottleMillis: 100
     # Allow search within same team only. Default: false
     # setSearchSameTeamOnly: false|true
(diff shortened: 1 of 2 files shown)


charts/brig/values.yaml
```diff
@@ -69,6 +69,8 @@ config:
     setDeleteThrottleMillis: 100
     # Allow search within same team only. Default: false
     # setSearchSameTeamOnly: false|true
+    # Allow settings max number of user clients. Default: 7
```
mheinzel:
```suggestion
    # Set max number of user clients. Default: 7
```
Slightly weird wording, this flag does not really allow setting a max number, but just sets it, right?

tiago-loureiro: Yes, thank you!

PR #189, opened 2020-02-20
Use brig-index to create index in ES

akshaymankar:
Fixes zinfra/backend-issues#1142.
Merge after wireapp/wire-server#964 is merged.


charts/elasticsearch-index/values.yaml
```diff
@@ -8,3 +8,6 @@ hook:
 # either a 'post'-subchart to an elasticsearch chart or a 'pre'-subchart to wire-server
   enabled: true
   type: pre # possible values: 'pre' or 'post'
+image:
+  repository: quay.io/wire/brig-index
+  tag: 2.69.0
```
akshaymankar: This is not the latest version, but I chose this to match versions of other images on `develop`.

tiago-loureiro: This version will actually not work, will it? As in, the 2.69 will fail with the above config?

akshaymankar: Hmm, yes it will not work. I guess we'll have to make a commit with all versions bumped after the new docker images are generated.

akshaymankar:
This now only depends on https://github.com/wireapp/wire-server/pull/1023, Once it is merged, we can add the correct version in values.yaml.

charts/elasticsearch-index/values.yaml
akshaymankar: After https://github.com/wireapp/wire-server/pull/1023 got merged, this works with brig-index-2.77.3. So, this PR is now ready to be reviewed and merged.

charts/elasticsearch-index/templates/job.yaml
```diff
@@ -31,16 +31,13 @@ spec:
...
-            - mountPath: /tmp/templates
-              name: es-init-volume
-
-      volumes:
-        - name: es-init-volume
-          configMap:
-            name: es-init-config
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          args:
+            - create
```
jschaul: Can you add a comment here that `create` actually means `create-if-not-exists`? It's not obvious from reading these arguments.

PR #190, opened 2020-02-24
Add team-id to brig-index

akshaymankar:
This is required because #189 and wireapp/wire-server#964 won't pass the tests
without each other. This commit should be mergeable without any changes in
wire-server and it will also make wire-server changes mergeable, which in turn
will let tests pass with #189.

Fixes zinfra/backend-issues#1142


akshaymankar:
Fixes zinfra/backend-issues#1142

akshaymankar:
This is not required any more, as https://github.com/wireapp/wire-server/pull/1023 has unblocked #189.

PR #197, opened 2020-03-07
Move hardcoded AWS_REGION env var value into chart values file

kvaps:


lucendio:
hey @kvaps,

thanks for the contribution. In order to make this work, I would like to ask you to adjust your change in a way so that the path to `region` is not broken. Meaning, in the templates `.Values.aws.region` has to be changed to `.Values.config.aws.region`, since the access is not wrapped in a dedicated context (e.g. `{{- with .Values.config }}...{{- end }}`).

kvaps:
sure, fixed

PR #200, opened 2020-03-13
Use apps/v1 for all deployments

akshaymankar:
These are all the cases I found. But I haven't tested this on a 1.16 cluster as I don't have one yet. 

We need to still make sure that all the helm charts we depend on are also compatible with K8s 1.16. Easiest way would be to try to deploy to K8s 1.16 and run the tests.

Fixes zinfra/backend-issues#1210

### DIFF
--- a/charts/account-pages/Chart.yaml
+++ b/charts/account-pages/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the Wire account pages in Kubernetes
 name: account-pages
-version: 0.89.0
+version: 0.89.0-pr.200
--- a/charts/account-pages/templates/deployment.yaml
+++ b/charts/account-pages/templates/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: account-pages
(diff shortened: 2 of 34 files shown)


akshaymankar:
zinfra/backend-issues#1210

akshaymankar:
This change requires at least K8s 1.9, which is ancient (released in Dec 2017). I don't think we need to specify this. I would be surprised if other things in our repo worked with K8s 1.9.

jschaul:
> This change requires at least K8s 1.9, which is ancient (released in Dec 2017). I don't think we need to specify this. I would be surprised if other things in our repo worked with K8s 1.9.

Ah ok. It seems some other things need to be changed though to make this PR work:

> release test-nfqvvp8nepzl-fake-aws failed: Deployment.apps "fake-aws-dynamodb" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string{"app":"fake-aws-dynamodb"}: `selector` does not match template `labels`]

akshaymankar:
The `apps/v1` API mandates selector and it must match the template labels. I've added it wherever it was missing. This does not break backwards compatibility though.

akshaymankar:
I am going to close and reopen this PR as I am sure it will fix the pipeline. I have created an issue for this here: https://github.com/zinfra/backend-issues/issues/1220

PR #203, opened 2020-03-17
Upgrade to external deps to be compatible with K8s 1.17

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1210

### DIFF
--- a/charts/redis-ephemeral/requirements.yaml
+++ b/charts/redis-ephemeral/requirements.yaml
@@ -1,5 +1,5 @@
 dependencies:
 - name: redis
-  version: 1.1.7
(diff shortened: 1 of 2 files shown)


jschaul:
This is in a draft state - why is that so?

akshaymankar:
> This is in a draft state - why is that so?

This change requires changes to cailleach and we deprioritized the issue, so we just haven't found time for this yet.

arianvp:
I'm going to close this. I have a redis bump working in the `nix-ansible` branch which I'll cherry-pick out of it and open a PR shorttly

PR #204, opened 2020-03-18
avs-nwtesttool / calling-test

jschaul:
Related to https://github.com/zinfra/backend-issues/issues/892

Obsoletes https://github.com/wireapp/wire-server-deploy/pull/160 for this purpose. That PR's jsonnet approach is still worth considering going forward, but requires quite a bit more work to integrate with existing systems.

* Makes [avs-nwtesttool](https://github.com/wireapp/wire-avs-nwtesttool) available as a helm chart (named `calling-test`).
* Adds a route to nginz under `/calling-test`, meaning the tool can be accessed under `https://https-nginz.example.com/calling-test` when both wire-server and the `calling-test` charts are deployed. A [small change](https://github.com/wireapp/wire-avs-nwtesttool/commit/4e771df61c66112d8bc95681079a6bebeb8f6390) was made to its config to support that route.

If we want, we can also add this chart as a subchart to the wire-server chart by default, what do you think? (I'll make that a separate PR though).


lucendio:
otherwise, lookks good to me

charts/calling-test/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+version: 0.1.0
```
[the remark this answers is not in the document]
jschaul: Yes. Once merged to develop, this will be published as the same version to the chart repo. Once merged to master, that will also be reflected in git. No manual version bumping needs to happen.

charts/calling-test/templates/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+          image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+        {{- range $key, $val := .Values.envVars }}
+            - name: {{ $key }}
+              value: {{ $val | quote }}
+        {{- end }}
+          ports:
+            - name: http
+              containerPort: 80
```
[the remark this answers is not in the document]
jschaul: I updated the image to use port 8080 (like we do in other charts).

jschaul:
> otherwise, lookks good to me

Please use the github buttons for approval if you approve, thanks.

charts/calling-test/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+image:
+  # note: the docker image tag is configured as 'appVersion' in Chart.yaml
+  repository: quay.io/wire/avs-nwtesttool
+  pullPolicy: IfNotPresent
+
+envVars:
+  # note: this should be overridden in every deployment
+  BACKEND_HTTPS_URL: https://nginz-https.example.com
+
+nameOverride: "calling-test"
```
tiago-loureiro: If someone overrides this name to something else, than the routing on the `nginz` level will no longer work, right? Maybe add a note along the lines of "Do not change this name unless you know what you are doing, routing over nginz will not work"?

charts/calling-test/templates/NOTES.txt
```diff
@@ -0,0 +1,5 @@
+1. Get the application URL by running these commands:
+
+  export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "calling-test.name" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
+  echo "Visit http://127.0.0.1:8080 to use your application"
+  kubectl --namespace {{ .Release.Namespace }} port-forward $POD_NAME 8080:8080
```
tiago-loureiro: Is this needed given that the application is available over `https://BACKEND_URL/calling-test`?

jschaul: It's only available that way when wire-server is installed. If installing this chart on your own (and e.g. pointing to a completely different server), it's still a useful way of testing that it works without needing full wire-server and nginz.

PR #205, opened 2020-03-19
use new override environment variable

jschaul:
(first requires new version of team settings to have an effect)


julialongtin:
Is there information about what change in team settings needs to be merged in order for us to merge this?

jschaul:
https://github.com/wireapp/wire-web-config-default/blob/master/wire-team-settings/.env.defaults#L23 has this label by default. Adding this environment variable override "too early" doesn't have any negative impact (just no effect). But this has been added in https://github.com/wireapp/wire-team-settings/pull/3264 so the recent releases should include this override possibility.

jschaul:
@julialongtin any further concerns?

julialongtin (approved):

PR #206, opened 2020-03-19
Remove deprecated option

tiago-loureiro:
As can be seen in the configmap, this setting is now long gone.


jschaul:
Already done in #202

PR #207, opened 2020-03-19
Remove deprecated option

tiago-loureiro:
Same as https://github.com/wireapp/wire-server-deploy/pull/206


jschaul:
Already done in #202

PR #211, opened 2020-03-25
Pin openjdk 8

arianvp:


jschaul:
There is also https://github.com/wireapp/ansible-cassandra/issues/10 that is related.

PR #213, opened 2020-04-07
Allow network interfaces not being set in inventory for backing services

lucendio:
* this change only reflects what is already been claimed in the docs: *Set the network interface name [...] if you have more than one network interface* (docs/how-to/install/ansible-VMs.rst:L123)
* unfortunately, Ansible does not allow reassigning a variable that can be undefined. In that case, Ansible throws and complains about the assigned variable (e.g.`elasticsearch_network_interface`) being undefined. Which makes it necessary do use the `default` filter when including the tasks and handing over variables.
* in the ES playbook `es_network_interface` var is already set at the top and takes proper defaulting into account. It only makes sense to use this value when iterating over master nodes

### DIFF
--- a/ansible/helm_external.yml
+++ b/ansible/helm_external.yml
@@ -2,8 +2,6 @@
 # that these databases listen on. These files are used as overrides with the
 # <database>-external helm charts (e.g. cassandra-external).
 #
-# Prerequisite: the '<database>_network_interface' must be defined.
-#
 # After any change to IPs/servers:
 # 1. run this playbook:
 #      poetry run ansible-playbook -i hosts.ini helm_external.yml -vv --diff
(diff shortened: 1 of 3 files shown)


ansible/helm_external.yml
```diff
@@ -24,25 +22,25 @@
       vars:
         external_dir_name: elasticsearch-external
         server_type: elasticsearch
-        network_interface: "{{ elasticsearch_network_interface }}"
+        network_interface: "{{ elasticsearch_network_interface | default('') }}"
```
jschaul, line 25: Defaulting to empty string is strange. If something is not set, we should not instead set it to an empty string: use the `omit` filter to keep `network_interface` undefined, which is easier to work with than empty string variables.

```suggestion
        network_interface: "{{ elasticsearch_network_interface | default(omit) }}"
```

(same for the other cases in this file)

ansible/templates/helm_external.yaml.j2
```diff
@@ -1,4 +1,4 @@
 IPs:
 {% for host in groups[server_type] %}
-  - {{ hostvars[host]["ansible_" + network_interface]["ipv4"]["address"] }}
+  - {{ hostvars[host]["ansible_" + (network_interface | default(hostvars[host].ansible_default_ipv4.interface, true)) ]["ipv4"]["address"] }}
```
jschaul, line 3: If `omit` is used above, then this parameter can be removed.

```suggestion
  - {{ hostvars[host]["ansible_" + (network_interface | default(hostvars[host].ansible_default_ipv4.interface)) ]["ipv4"]["address"] }}
```

lucendio:
@jschaul regarding your suggestion: unfortunately this wont work. 
```
TASK [write IPs for helm] **********************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'ansible___omit_place_holder__d63f9436bb99cb21faeb8998c9394fd302aacdfb'"}
```
`omit` just seems to be yet another global var holding a string value, which is not respected by another `default` filter.

I banged my head against the wall trying to come up with sth cleaner than this. Apparently, Ansible is very strange when it comes to reassigning variables whose value is undefined. I tried your suggestion and various other ways. E.g, I also suspected *include* and *import* - even though it should be the preferred way - but it doesnt make any difference.

Happy to be proven wrong here. Maybe I overlooked sth.

PR #214, opened 2020-04-08
configure maxScimTokens setting

orandev:
maxScimTokens value in the chart values file is 0, which prevents the creation of Scim Tokens
I propose to set it to 16 in the values example file.

### DIFF
--- a/values/wire-server/prod-values.example.yaml
+++ b/values/wire-server/prod-values.example.yaml
@@ -157,6 +157,7 @@ spar:
     ssoUri: https://nginz-https.example.com/sso
     maxttlAuthreq: 28800
     maxttlAuthresp: 28800
+    maxScimTokens: 16
     contacts:
     - type: ContactSupport
       company: YourCompany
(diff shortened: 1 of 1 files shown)


fisx:
I think the default of 0 was chosen as a way of turning the feature off by default: people who don't want to use SCIM can be confident it won't be used, not even accidentally.

orandev:
Hello,
Ok, I have modified the code to make it a comment, for people like me to know how to make SCIM work without having to deeply search into the source code.

fisx:
> Hello,
> Ok, I have modified the code to make it a comment, for people like me to know how to make SCIM work without having to deeply search into the source code.

Perfect, thanks!  We'll think of that next time.

lucendio:
@orandev I'm curious, which use case or requirement made you choose to set it to `16`?

@fisx @jschaul 
How about splitting the purpose of that variable into two:
(1) a bool flag to en-/disable
(2) a number (with a reasonable default) to set the max amount of tokens?

orandev:
Hello
[docs/reference/provisioning/scim-token.md](https://github.com/wireapp/wire-server/blob/develop/docs/reference/provisioning/scim-token.md)
"There is a reasonable limit on the number of tokens a single team can have, set in `scim.yaml` at `maxScimTokens`. For Wire the limit is 16."

PR #217, opened 2020-04-09
configure proxy env vars in some charts

orandev:
In some enterprise environments, a proxy is needed to access the Internet.
Some pods need to access the Internet to access AWS, a cloud Object Storage provider if not using local Minio instance, Giphy, etc...
This change permits admins to configure http_proxy, https_proxy and no_proxy env vars inside these pods, from the wire-server values yaml file.

### DIFF
--- a/charts/brig/templates/deployment.yaml
+++ b/charts/brig/templates/deployment.yaml
@@ -66,24 +66,26 @@ spec:
           # TODO: Is this the best way to do this?
           - name: AWS_REGION
             value: "{{ .Values.config.aws.region }}"
-          {{- if .Values.config.proxy.httpProxy }}
+          {{- with .Values.config.proxy }}
+          {{- if .httpProxy }}
           - name: http_proxy
-            value: {{ .Values.config.proxy.httpProxy | quote }}
+            value: {{ .httpProxy | quote }}
           - name: HTTP_PROXY
-            value: {{ .Values.config.proxy.httpProxy | quote }}
+            value: {{ .httpProxy | quote }}
           {{- end }}
-          {{- if .Values.config.proxy.httpsProxy }}
+          {{- if .httpsProxy }}
           - name: https_proxy
-            value: {{ .Values.config.proxy.httpsProxy | quote }}
+            value: {{ .httpsProxy | quote }}
           - name: HTTPS_PROXY
-            value: {{ .Values.config.proxy.httpsProxy | quote }}
+            value: {{ .httpsProxy | quote }}
           {{- end }}
-          {{- if .Values.config.proxy.noProxy }}
+          {{- if .noProxyList }}
           - name: no_proxy
-            value: {{ .Values.config.proxy.noProxy | quote }}
+            value: {{ join "," .noProxyList | quote }}
           - name: NO_PROXY
-            value: {{ .Values.config.proxy.noProxy | quote }}
-          {{- end }}      
+            value: {{ join "," .noProxyList | quote }}
+          {{- end }}
+          {{- end }} 
           ports:
             - containerPort: {{ .Values.service.internalPort }}
           livenessProbe:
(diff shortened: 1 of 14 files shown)


lucendio:
@orandev Thanks for the PR!

If I understand you correctly, you aim to address a specific use case with this PR, which is enabling to set an HTTP proxy in containers. Since this is usually done via environment variable, the solution presented here introduces the ability to set arbitrary environment variables.

All wire components are configured via config files. So, from a software design perspective, it is and will not be necessary to set arbitrary environments variables for any of them. When talking about containers and engines such as Docker, setting arbitrary environment variables might actually allow to leverage potential vulnerabilities, we might not know of (yet). To be honest, in part, it's a gut feeling.

I would suggest a more explicit approach. Meaning, to define Helm variables specifically for the proxy use case and propagate their values down to the sub-chart templates that might need them and set the environments variables `HTTPS_PROXY`, `HTTP_PROXY`and `NO_PROXY` there.
With this, It might be desired, that one defaults to the other, if not set.

Happy to hear your opinion on this.

orandev:
Hello,

"All wire components are configured via config files."
No, you allow admins to set arbitrary environments variables in webapp, team-settings and account charts. I just duplicated the code from these charts.

I can modify the code as you said if you prefer. But if somebody has the rights to install/update a chart on a K8S cluster, doesn't it mean that he has admin access to the K8S cluster, and thus can modify any env variable in any container anyway?

lucendio:
For the 3 fontend components (webapp, team-settings, account-pages), you are right, yes.
Their developers choose to follow a different path to configure them. The concern remains, though, and we can't do much about it atm. 

I understand your reasoning and why you made the changes in this fashion. My thinking is more around the perspective from inside a container towards the outside. Furthermore, I'd like to believe that trying to prevent/reduce any way of abuse is good practice.

So, adjusting your PR would be much appreciated. Please, feel free to share your thoughts on how this should be implemented.

orandev:
Hello,
I've modified the code, could you tell me if it's ok for you?

charts/brig/templates/deployment.yaml
```diff
@@ -66,6 +66,24 @@ spec:
...
+            value: {{ .Values.config.proxy.httpsProxy | quote }}
+          - name: HTTPS_PROXY
+            value: {{ .Values.config.proxy.httpsProxy | quote }}
+          {{- end }}
+          {{- if .Values.config.proxy.noProxy }}
+          - name: no_proxy
+            value: {{ .Values.config.proxy.noProxy | quote }}
+          - name: NO_PROXY
+            value: {{ .Values.config.proxy.noProxy | quote }}
+          {{- end }}
```
lucendio, line 86: Thank you for all your effort! Some minor nitpicking, but free to integrate it or not :)

In this case, setting the scope can simplify the overall code a bit:
```
{{- with .Values.config.proxy }}
{{- if .httpProxy }}
- name: http_proxy
  value: {{ .httpProxy | quote }}
- name: HTTP_PROXY
  value: {{ .httpProxy | quote }}
{{- end }}
{{- if .httpsProxy }}
- name: https_proxy
  value: {{ .httpsProxy | quote }}
- name: HTTPS_PROXY
  value: {{ .httpsProxy | quote }}
{{- end }}
{{- if .noProxyList }}
- name: no_proxy
  value: {{ join "," .noProxyList | quote }}
- name: NO_PROXY
  value: {{ join "," .noProxyList | quote }}
{{- end }}
{{- end }}
```

charts/brig/values.yaml
```diff
@@ -73,6 +73,10 @@ config:
     # setUserMaxPermClients: <int>
   smtp:
     passwordFile: /etc/wire/brig/secrets/smtp-password.txt
+  proxy:
```
lucendio, line 76: I think, it should be sufficient to just set `proxy: {}`

lucendio:
Thanks for your effort, very much appreciated!

PR #220, opened 2020-04-14
better defaults

jschaul:


tiago-loureiro:
Have you tried this out? Last time I tried this out, it caused the redirect to go over the HTTPS port on the container which doesn't exist/doesn't do the right thing since we do SSL termination at the ingress level (as it's done for other services) so all traffic arriving at the container comes in plain text.

orandev:
I have the 3 variables set to True since I installed my platform and everything seems to work fine.

julialongtin (approved):

PR #212, opened 2020-03-27
Add brig index migrations

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1142
Depends on https://github.com/wireapp/wire-server/pull/964

Summary of changes:
- The elasticsearch-index chart now mandatorily runs the job to create index pre-uprgade and pre-install
- The create index job also puts mapping
- There is a new post-upgrade and post-install job to run data-migrations on the index.

Deleted values:

- elasticsearch-index.hook.*

New values in the chart:
- elasticsearch-index.cassandra.host
- elasticsearch-index.cassandra.port
- elasticsearch-index.cassandra.keyspace

To do before merge:
- [x] Merge https://github.com/zinfra/cailleach/pull/329

PR #199, opened 2020-03-11
Verify proper NTP installation on Cassandra hosts

tiago-loureiro:
Too often we have been bitten by this; this adds a final check on the cassandra playbook that NTP is _indeed_ properly setup. Furthermore, it adds a simple playbook to verify cassandra hosts when necessary. Can be adapted/adjusted to run on all hosts, and potentially even added to https://github.com/geerlingguy/ansible-role-ntp

PR #221, opened 2020-04-15
release_2020_04_15

akshaymankar:
# Changelog

## Release Notes

- This version adds a new migration to the elasticsearch index, it will go through all users in
  cassandra and (re-)create all users in elasticsearch. So, it could take a long time to finish 
  depending on the number of users in the system.

## Features

- Use brig-index to create index in ES (#189)
- Allow docker registry to run with custom host and port (023eb5e)
- Verify proper NTP installation on Cassandra hosts (#199, c1acc03)
- Pin openjdk 8 (#211)
- Add brig index migrations (#212)
- Bump external role ansible-helm to support installing newer versions of Helm (c86c36f)
- Add ES, restund_network_interface and http_proxy settings to terraform inventory template (#216)
- Add important envVars for team-settings and account-pages in example values (#215)
- Add comment about configuring maxScimTokens setting (#214)
- Lock ES version in ES ansible (#219)
- Add comment about restund_network_interface in example hosts.ini (#219)
- Allow network interfaces being unset in inventory for nodes hosting backing services (#213)

## Bug fixes

- Fix incorrect bash used in docker-registry (cb73c38)


charts/elasticsearch-index/values.yaml
```diff
@@ -3,8 +3,10 @@ elasticsearch:
...
-# either a 'post'-subchart to an elasticsearch chart or a 'pre'-subchart to wire-server
-  enabled: true
-  type: pre # possible values: 'pre' or 'post'
+cassandra:
+  # host:
+  port: 9042
+  keyspace: brig
+image:
+  repository: quay.io/wire/brig-index
+  tag: 2.77.11
```
tiago-loureiro, line 12: We can now remove this tag, right?

akshaymankar: I think it would get overwritten when we merge wire-server. Until then, this is a working build

tiago-loureiro: Or maybe comment it out and leave a comment that: _it must be at least version 2.77.11_ ?

akshaymankar: Hmm, let me see if cailieach will like a comment there. I think we do some sed magic to bump versions. Why is it a problem if we have this version here?

akshaymankar: My bad, it wasn't caillieach, it was [set-version.sh](https://github.com/wireapp/wire-server-deploy/blob/3bc2db80c2cb05917c82333e53bbf7b0ba470878/bin/set-version.sh#L15) in this repo, we rely on that for the version to keep ticking. So, we can leave the make it `version: ` and leave a comment above. But I am not sure why we would go through the hassle.

akshaymankar:
> Can we add a comment stating that, using these charts+images, there will be an automated data migration so expect the first deployment to take some time as it needs to create all users from C* to ES?

Makes sense. I will add it to CHANGELOG and release notes in the description of this PR.

charts/elasticsearch-index/values.yaml
tiago-loureiro: > Hmm, let me see if cailieach will like a comment there. I think we do some sed magic to bump versions. Why is it a problem if we have this version here?

Hmmm, I think you are right. It will get overwritten in the CI process. I just wanted to make sure to _not_ pin upcoming charts (on develop and master) to this version.

PR #224, opened 2020-04-16
skip memorizing the IPs of redis nodes if there are not any.

julialongtin:


julialongtin:
this actually prevents ansible from failing, when the group redis does not exist.

julialongtin:
I thought of that, but if the list is empty, we want to remove IPs from the
associated files. or, at least, that's the least surprising behavior.

lucendio:
> I thought of that, but if the list is empty, we want to remove IPs from the associated files. or, at least, that's the least surprising behavior.

Speaking like a true idempotence wizard!

PR #229, opened 2020-04-21
added aws region in brig and galley in prod values example file

orandev:
Prod values example file was not updated after https://github.com/wireapp/wire-server-deploy/pull/197

### DIFF
--- a/values/wire-server/prod-values.example.yaml
+++ b/values/wire-server/prod-values.example.yaml
@@ -28,6 +28,7 @@ brig:
     useSES: false
     aws:
       # change if using real AWS
+      region: "eu-west-1"
       sqsEndpoint: http://fake-aws-sqs:4568
       dynamoDBEndpoint: http://fake-aws-dynamodb:4567
       # these must match the table names created on fake or real AWS services
@@ -110,6 +111,8 @@ galley:
     settings:
       # prefix URI used when inviting users to a conversation by link
       conversationCodeURI: https://webapp.example.com/join/ # change this
+    aws:
+      region: "eu-west-1"
 
 gundeck:
   replicaCount: 3
@@ -122,7 +125,7 @@ gundeck:
     aws:
       # change if using real AWS
       account: "123456789012"
-      region: eu-west-1
+      region: "eu-west-1"
       arnEnv: integration
       queueName: integration-gundeck-events
       sqsEndpoint: http://fake-aws-sqs:4568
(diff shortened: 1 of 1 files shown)


orandev:
You force the region in the fake aws charts:
_aws configure set region eu-west-1_
Shouldn't it be the same region in the prod values example file since it is fake-aws that is configured there?

lucendio:
That is a good point, but `fake-aws` is somewhat independent from the example values files. Those files are meant to override defaults not defining them. 
Marking such a value as suggested above would explicitly point out that it needs review or at least should be questioned.

That said, in terms of `region` in `fake-aws`, it needs refactoring. In recent PRs it was started to make `region` more parameterize and less hard-coded. We need to pursue this path all the way to the end.

julialongtin:
@lucendio can you link the issue that we need to solve before this can be merged / rejected?

lucendio:
__tl;dr:__ I think we can merge this PR with a minor adjustment: @orandev can you please rollback to your [first commit](https://github.com/wireapp/wire-server-deploy/pull/229/commits/2f04ca9a3908f3b9c654017a5e477cb6571fc516). 

 #### Long story

The question, we need to ask here is, why are certain values set in the example values files and others are not? Well, when deploying Wire on Kubernetes the default assumption is an on-prem or private cloud installation - in any case, a scenario that is far away from AWS. 

Galley, Brig and Gundeck have all feasible default values for this scenario (fake-aws). But at the same time, the examples try to indicate where things need to be adjusted (`# change if using real AWS`), in case it a really runs on AWS. But sadly it's incomplete and inconsistent.

This change is a right step to address the latter. Follow-up work is tracked [here](https://github.com/zinfra/backend-issues/issues/1421)

orandev:
> the default assumption is an on-prem or private cloud installation - in any case, a scenario that is far away from AW

We were advised to use real AWS on our on-prem platform so that is incorrect.

lucendio:
> > the default assumption is an on-prem or private cloud installation - in any case, a scenario that is far away from AWS
> 
> We were advised to use real AWS on our on-prem platform so that is incorrect.

Of course, use cases may vary. I was referring to the __default__ perspective, when looking at the examples, and reference implementation that we are referring to in our [deployment guide](https://docs.wire.com/how-to/index.html) and aims to be cloud-agnostic.

One last request. Would you mind rebasing the branch against develop. after that, I'll will merge it!

jschaul:
(we may wish to take multiple perspectives at some point: one being a production on-premise, and one being a production on-aws deployment. This requires that we first find a way to not have all of these value files drift apart so much though. Maybe we can make use of some templating. Or some abstraction around these values)

PR #230, opened 2020-04-21
add terraform configuration from the offline environment.

julialongtin:


terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+    region  = "eu-central-1"
+
+    # TODO: create IAM policy which only allows access to this bucket under
+    # envrionments/crash and to the -crash dynamodb table
+    bucket = "z-terraform-remote-state"
+
+    key = "environments/offline/bootstrap"
+
+    dynamodb_table = "z-terraform-state-lock-dynamo-lock-environment-offline"
+  }
```
jschaul, line 24: Remove the entire `backend "s3"` block (leaving only the version constraint) - it's not needed inside a module, as module users will define this, not the modules themselves

terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+}
+
+# Used for the in-VPC EC2 endpoint.
+data "aws_security_group" "default" {
+  name   = "default"
+  vpc_id = module.vpc.vpc_id
+}
+
+module "vpc" {
+  source = "github.com/terraform-aws-modules/terraform-aws-vpc"
```
jschaul, line 39: Can you version lock this? Example syntax: `source = "git::https://github.com/cloudposse/terraform-aws-s3-website.git?ref=tags/0.8.0"`

terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+    Owner       = "Backend Team"
+    Environment = "Offline"
+    TrustZone   = "Crash"
+  }
+  vpc_tags = {
+    Owner       = "Backend Team"
+    Name        = "vpc-offline"
+    TrustZone   = "Crash"
+  }
+}
```
jschaul, line 81: This block (from "provider aws" until the end of vpc) should be moved to `terraform/modules/aws_vpc`. A few things need to be put into variables. So the domain name, the environment, and the trustZone are all good candidates for variables. Declaring and using variables is described [on this page](https://www.terraform.io/docs/configuration/variables.html). I wouldn't recommend to over-do variables at this point, so in a first stage I think we can keep the subnets and the availability zones hard-coded. (Or we use variables with the existing values as a [default](https://www.terraform.io/docs/configuration/variables.html#declaring-an-input-variable))

terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+
+# A SSH key, used during golden image creation. destroyed at the end of the process.
+resource "aws_key_pair" "crash-nonprod-deployer-julia" {
```
jschaul, line 84: I think this (aws key) can be removed entirely from this location.

terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+    values = ["ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-*"]
+  }
+
+  filter {
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  owners = ["099720109477"] # Canonical
+  }
```
jschaul, line 122: This block (for finding AMIs) could be deleted, or placed into `terraform/modules/aws_ami`.

terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+    protocol    = "tcp"
+    cidr_blocks = ["172.17.0.0/20"]
+  }  
+
+  tags = {
+    Name = "ephemeral_private"
+  }
+}
+
+# A security group for ephemeral service nodes. should be added to them only.
```
jschaul, line 487: What is an ephemeral service node?

terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+    from_port   = 9092
+    to_port     = 9092
+    protocol    = "tcp"
+    security_groups = ["${aws_security_group.ephemeral_private.id}"]
+  }
+
+  tags = {
+    Name = "ephemeral_node"
+  }
+}
```
jschaul, line 568: I'm not exactly sure what to do about security groups. I suppose we can, for the time being, keep these hardcoded ports as-is, and move all the `aws_security_group` blocks into `terraform/modules/aws_security_groups/main.tf`

terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+
+# Finding Instance types:
+# https://www.ec2instances.info/
+
+# point an elastic IP to our bastion host.
+resource "aws_eip" "bastion-offline" {
+  vpc                       = true
+#network_interface         = "${aws_network_interface.bastion-crash-in.id}"
+  instance = "${aws_instance.bastion-offline.id}"
+}
```
jschaul, line 132: The elastic IP block should be moved next to the instance creation code.

terraform/examples/offline.tf
```diff
@@ -0,0 +1,761 @@
...
+    security_groups = ["${aws_security_group.ephemeral_private.id}"]
+  }
+
+  tags = {
+    Name = "ephemeral_node"
+  }
+}
+
+# our bastion host
+resource "aws_instance" "bastion-offline" {
```
jschaul, line 571: I would suggest to remove all the `aws_instance` blocks from this PR entirely. I think it makes sense to define these on the user level (for us: inside an environment folder), not as a module, depending on requirements for that environment. We can discuss that separately on how best to proceed, and possibly create another PR with another module, if we deem that a good strategy.

PR #231, opened 2020-04-22
Add maxFanout to galley's options

tiago-loureiro:
This will be required for https://github.com/wireapp/wire-server/pull/1060 to pass the tests

### DIFF
--- a/charts/galley/templates/configmap.yaml
+++ b/charts/galley/templates/configmap.yaml
@@ -43,7 +43,9 @@ data:
       intraListing: false
       maxTeamSize: {{ .settings.maxTeamSize }}
       maxConvSize: {{ .settings.maxConvSize }}
+      {{- if .settings.truncationLimit }}
       truncationLimit: {{ .settings.truncationLimit }}
+      {{- end }}
       conversationCodeURI: {{ .settings.conversationCodeURI | quote }}
       {{- if .settings.featureFlags }}
       featureFlags:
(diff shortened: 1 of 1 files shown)


charts/galley/templates/configmap.yaml
```diff
@@ -43,6 +43,7 @@ data:
       intraListing: false
       maxTeamSize: {{ .settings.maxTeamSize }}
       maxConvSize: {{ .settings.maxConvSize }}
+      truncationLimit: {{ .settings.truncationLimit }}
```
akshaymankar, line 46: It will make upgrades easier if we default this to 500 like the `maxTeamSize`.

tiago-loureiro: Actually this is optional, and in general nobody should set it. Maybe we can adjust it to:
```
{{- if .settings. truncationLimit }}
    truncationLimit: {{ .settings.truncationLimit }}
{{- end }}
```

akshaymankar: Yeah, that is even better.

tiago-loureiro: https://github.com/wireapp/wire-server-deploy/pull/231/commits/2c4a00d6854a48074f812fa521ffe81351522f86

tiago-loureiro:
`maxFanoutSize` is indeed a better name

PR #236, opened 2020-04-23
move to helm 3.

julialongtin:
grab helm 3, do not intitalize helm, and do not tell kubespray to deploy helm2 on the nodes.

### DIFF
--- a/ansible/download_cli_binaries.yml
+++ b/ansible/download_cli_binaries.yml
@@ -11,7 +11,7 @@
   vars:
     # Ensure these versions match the server-side ones (or are compatible)
     # Also see download_kubespray.yml
-    kubernetes_helm_ver: v3.1.0
+    kubernetes_helm_ver: v3.1.1
     kubectl_version: "1.14.2"
     # default directories for a global installation
     kubectl_bin_directory: "/usr/local/bin"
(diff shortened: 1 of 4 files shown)


ansible/download_cli_binaries.yml
```diff
@@ -11,7 +11,7 @@
   vars:
     # Ensure these versions match the server-side ones (or are compatible)
     # Also see download_kubespray.yml
-    kubernetes_helm_ver: v2.13.1
+    kubernetes_helm_ver: v3.1.0
```
jschaul, line 14: Can we make this:

```suggestion
    kubernetes_helm_ver: v3.1.1
```

ansible/hosts.example.ini
```diff
@@ -145,8 +145,6 @@ bootstrap_os = ubuntu
 # docker_dns_servers_strict = false
 
 [k8s-cluster:vars]
-# gotcha: use capital True or False (otherwise ansible parses as string)
-helm_enabled = True
```
jschaul, line 149: this also needs to be removed from:

```
ansible/hosts.example-demo.ini
43:helm_enabled = True

terraform/examples/inventory.tpl
94:helm_enabled = True
```

PR #241, opened 2020-04-24
Add module to enable mobile push notification (used by Gundeck)

lucendio:


terraform/modules/aws_gundeck-push-notifications/resources.sns.tf
```diff
@@ -0,0 +1,54 @@
...
+  # ^-- Path to the secret token
+  event_delivery_failure_topic_arn = "arn:aws:sns:${var.region}:${var.account_id}:${var.environment}-${var.queue_name}"
+  # ^-- Topic to subscribe to
+  event_endpoint_updated_topic_arn = "arn:aws:sns:${var.region}:${var.account_id}:${var.environment}-${var.queue_name}"
+  # ^-- Topic to subscribe to
+}
+
+# Create topics and queues to publish push notifications
+
+resource "aws_sns_topic" "incoming_message_or_call" {
```
tiago-loureiro, line 45: This topic is not related to notifications; it's for feedback from Apple/Google that some device/state has changed.

tiago-loureiro: Can we rename it to something like `notification_feedback`?

terraform/modules/aws_gundeck-push-notifications/outputs.tf
```diff
@@ -0,0 +1,22 @@
+# Output required to configured wire-server
+
```
tiago-loureiro, line 2: Can we add:

 * "arnEnv" to the output? Might not be obvious

terraform/modules/aws_gundeck-push-notifications/resources.iam.tf
```diff
@@ -0,0 +1,50 @@
+resource "aws_iam_user" "gundeck" {
+  name          = "${var.environment}-gundeck-full-access"
+  force_destroy = true
```
tiago-loureiro, line 3: This is not idempotent, I am not sure we should `force_destroy` (if there's a running gundeck instance it would lose access to AWS). I think we should just fail hard if this fails.

terraform/modules/aws_gundeck-push-notifications/resources.sns.tf
```diff
@@ -0,0 +1,54 @@
...
+  # to register for push notifications
+  # More details: https://github.com/zinfra/backend-wiki/wiki/Native-Push-Notifications#ios
+  #
+  platform = "APNS_VOIP"
+  # ^-- We only use VoIP at the moment
+  platform_principal = file("${var.apns_credentials_path}.cert.pem")
+  # ^-- Path to the public certificate
+  platform_credential = file("${var.apns_credentials_path}.key.pem")
+  # ^-- Path to the private key
+  event_delivery_failure_topic_arn = "arn:aws:sns:${var.region}:${var.account_id}:${var.environment}-${var.queue_name}"
```
tiago-loureiro, line 19: Oh I think this was overlooked from my end on my original PR, can we change this to:
```suggestion
  event_delivery_failure_topic_arn = "${aws_sns_topic.incoming_message_or_call.arn}"
```

terraform/modules/aws_gundeck-push-notifications/variables.tf
```diff
@@ -0,0 +1,45 @@
...
+  default     = "eu-central-1"
+}
+
+variable "environment" {
+  type        = string
+  description = "name of the environment as a scope for the created resources (e.g., prod. Can be staging or anything else)"
+  default     = "dev"
+}
+
+variable "account_id" {
```
tiago-loureiro, line 13: Why do we want this to be a parameter rather than using the account ID associated with the caller?

terraform/modules/aws_gundeck-push-notifications/variables.tf
```diff
@@ -0,0 +1,45 @@
...
+  type        = string
+  description = "AWS account the created resourcces are need to be assigned to"
```
tiago-loureiro, line 15:
```suggestion
  description = "AWS account the created resources are need to be assigned to"
```

terraform/modules/aws_gundeck-push-notifications/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+```hcl-terraform
+module "gundeck-push-notification" {
+  source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws_gundeck-push-notifications?ref=develop"
+  
+  region = "eu-central-1"
+  environment = "dev"
+  account_id = ""
+  apns_application_id = "wire.com"
+  apns_credentials_path = "${path.root}/path/to/app-credentials/files"   # omit '.[cert,key].pem'
```
tiago-loureiro, line 17: Can we assume that the files will be named `voip`? This shows that it's a `voip` certificate (there are other types but we are not using them currently
```suggestion
  apns_credentials_path = "${path.root}/path/to/app-credentials/files"   # omit 'voip.[cert,key].pem'
```

terraform/modules/aws_gundeck-push-notifications/resources.iam.tf
lucendio: I think the naming of that field is a bit misleading, take a look at the [docs](https://www.terraform.io/docs/providers/aws/r/iam_user.html#force_destroy). Should be fine IMHO,, though, i just c/p that from *engagements*

terraform/modules/aws_gundeck-push-notifications/outputs.tf
lucendio: From what I've see, `arnEnv == var.environment`. Though, `environment` is not a required variable. So, I guess, it makes sense output this as well. Nevertheless, I feel a bit reluctant to naming it `ancEnv`, because that is actually - or at least for me - confusing. And IMHO it is indeed just the name of the environment, which we use for prefixing. 
I think, the issue here is that the variables of the chart values are called that way. A more general name for this is probably more fitting, since other technologies will be involved in the future, too.

What do you think?

terraform/modules/aws_gundeck-push-notifications/resources.sns.tf
lucendio: renamed it to `device_state_changed`. I hope that is fine.

terraform/modules/aws_gundeck-push-notifications/resources.sns.tf
lucendio: same for `event_endpoint_updated_topic_arn` two LOCs below?

terraform/modules/aws_gundeck-push-notifications/variables.tf
lucendio: When we modularize this logic, I think that assumption is no longer valid.

terraform/modules/aws_gundeck-push-notifications/README.md
lucendio: Hard-coding this, I think, has no benefit, but also doesnt prevent you form naming the files that way. I'm even hesitant to use the path to the files as variable. I'd would rather have the file content itself to be the variable value. I only decided otherwise, because then we would need another var for APNS, since it requires to files for credentials.
But the more I'm thinking about it, the more I'm open to this, because it's more explicit and visible.

terraform/modules/aws_gundeck-push-notifications/resources.sns.tf
tiago-loureiro: Yes, everywhere 👍

terraform/modules/aws_gundeck-push-notifications/variables.tf
tiago-loureiro: How do u envision that being used? The caller would create resources on different accounts?

terraform/modules/aws_gundeck-push-notifications/outputs.tf
tiago-loureiro: `environment` is not required because there's always a default, right? That's fine, it must be present so the default sounds sensible.

terraform/modules/aws_gundeck-push-notifications/resources.iam.tf
tiago-loureiro: > I think the naming of that field is a bit misleading, take a look at the docs. Should be fine IMHO

True, I am confused.

> though, i just c/p that from engagements

Yup, well aware. I think it did recreate the user, I can re-test it 🤔

terraform/modules/aws_gundeck-push-notifications/README.md
tiago-loureiro: I think hardcoding and forcing some convention is fine; if not, then we should at least make it explicit what sort of credentials we are looking for.

terraform/modules/aws_gundeck-push-notifications/outputs.tf
lucendio: yes. right now we follow this approach with all modules. unfortunately for AWS, everything under one account shares a naming scope. and, as you are aware of, some things are even platform global ... m)

terraform/modules/aws_gundeck-push-notifications/resources.iam.tf
lucendio: Yes, please do. For now I keep it in there. It's used in different modules. But for each case, we create a user, there is also an access key created. So, at least for those the rule from the docs does not apply. So, I think we could safely omit this flag.

Let me know your findings when re-testing this.

terraform/modules/aws_gundeck-push-notifications/variables.tf
lucendio: Thx for bumping my nose into this. It appears, I had sth else in my head, that I compared to a concept in GCP. You are right, it doesnt really make sense.
Though, since we changed `event_delivery_failure_topic_arn` and alike, at least for the gundeck module, we dont need the account ID anymore. That said, since it is still required for gundeck's helm configuration, I suggest, we put the data resource to obtain the account ID somewhere in the TF environment code to output it from there  and be able to set the helm values correctly.

terraform/modules/aws_gundeck-push-notifications/README.md
lucendio: I decided to go with providing file content instead of the path and ended up with 3 variables overall (2 apns, 1 gcm). the readme shows how to use it. Therefore, the modules does not impose any file naming. It is up to the user how to organize their files.

smoesai:
How do I get this push notification module working? Do I have to copy this:

module "gundeck-push-notification" {
  source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-gundeck-push-notifications?ref=CHANGE-ME"

  environment = "dev"
  apns_application_id = "myapp.tld"
  apns_voip_key = file("path/to/app-credentials/key.pem")
  apns_voip_cert = file("path/to/app-credentials/cert.pem")
  gcm_application_id = "123456789"
  gcm_key = file("path/to/app-credentials/key.txt")
}

in the wire-server/values.yaml in the gundeck section?

jschaul:
> How do I get this push notification module working? Do I have to copy this:
> 
> module "gundeck-push-notification" {
> source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-gundeck-push-notifications?ref=CHANGE-ME"
> 
> environment = "dev"
> apns_application_id = "myapp.tld"
> apns_voip_key = file("path/to/app-credentials/key.pem")
> apns_voip_cert = file("path/to/app-credentials/cert.pem")
> gcm_application_id = "123456789"
> gcm_key = file("path/to/app-credentials/key.txt")
> }
> 
> in the wire-server/values.yaml in the gundeck section?
> 
> And is this the way to NOT use AWS SNS and SQS?

This module can be used if you want to use real SNS and real SQS for real push notifications using the official Wire mobile clients from the appstore and Playstore. However, you need to enter into an agreement with Wire for that, see also https://docs.wire.com/how-to/install/configuration-options.html#enable-push-notifications-using-the-public-appstore-playstore-mobile-wire-clients

If you are fine with websocket-only notifications, you don't need this module at all.

smoesai:
Notifications over websockets is not working on my own production server. Everything else is working fine. Gundeck and Cannon have the default settings. I am not receiving any message notification when the app is in the background. Please help me..

jschaul:
On android, settings/advanced/ "maintain connection to server":

maintain

smoesai:
I don’t have that option in iOS.. is there another way to get push notifications on iOS without using AWS?

jschaul:
> I don’t have that option in iOS.. is there another way to get push notifications on iOS without using AWS?

No. You can either not use push notifications on iOS (it makes iOS app rather unusable), or get in touch with Wire over the support/sales channels to get an AWS SNS/SQS setup.

smoesai:
@jschaul thanks for your time and the information. I will contact Wire Support/Sales

smoesai:
I tried multiple to get in touch with Wire support/sales, but I am not getting any response from them.. Is there any other way to contact them?

jschaul:
> I tried multiple times to get in touch with Wire support/sales, but I am not getting any response from them.. Is there any other way to contact them?

How did you contact them?

smoesai:
Through these contact forms:

https://wire.com/en/contact/sales/

https://support.wire.com/hc/en-us/requests/new?ticket_form_id=101635

smoesai:
Is there another way?

jschaul:
Those are correct channels. It can sometimes take some time until someone from support or sales gets the time to answer requests. The "on premise" things are generally handled under the "enterprise" tier, where the contact form is https://wire.com/en/contact/enterprise-requests/

smoesai:
Thank you for your help

PR #243, opened 2020-04-27
Add module to create a bucket in an object storage (S3) on AWS (used by Cargohold)

lucendio:


terraform/modules/aws_cargohold-asset-storage/README.md
```diff
@@ -0,0 +1,20 @@
...
+=========================================
+
+This module creates an Object Storage on AWS for cargohold to store encrypted assets.
+
+
+#### How to use the module
+
+```hcl-terraform
+module "gundeck-push-notification" {
+  source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws_cargohold-asset-storage?ref=develop"
```
jschaul, line 11: module name doesn't match the source (name is about gundeck, source abour cargohold). Rename?

Also, could we agree on all-underscores? It's slightly strange to have an underscore after aws but not between the other words. I'm not very opinionated about `_` or `-`, just I'd like to use them consistently, and not a mixture.

terraform/modules/aws_cargohold-asset-storage/variables.tf
```diff
@@ -0,0 +1,21 @@
...
+  default     = "eu-central-1"
+}
+
+variable "environment" {
+  type        = string
+  description = "name of the environment as a scope for the created resources (e.g., prod. Can be staging or anything else)"
+  default     = "dev"
+}
+
+variable "account_id" {
```
jschaul, line 13: this variable is never used.

terraform/modules/aws_cargohold-asset-storage/variables.tf
```diff
@@ -0,0 +1,21 @@
+variable "region" {
```
jschaul, line 1: this variable is never used

terraform/modules/aws_cargohold-asset-storage/variables.tf
```diff
@@ -0,0 +1,21 @@
...
+}
+
+variable "account_id" {
+  type        = string
+  description = "AWS account the created resources are need to be assigned to"
+}
+
+variable "bucket_name" {
+  type = string
+  description = "Name of the bucket that cargohold uses to store files, e.g. 'prod-assets'"
```
jschaul, line 20: add `default = "assets"` as that name is what is in use in helm charts as well. The comment should mention that the real bucket name will be prefixed with the environment. (The existing comment is misleading here)

terraform/modules/aws_cargohold-asset-storage/README.md
```diff
@@ -0,0 +1,20 @@
...
+  
+  region = "eu-central-1"
+  environment = "staging"
+  account_id = "myAccountID"
+  bucket_name = "my-assets"   # will be prefixed with $environment
```
jschaul, line 16: Leave out bucket_name, account_id, and region, as those are not used and/or can have a reasonable default.

terraform/modules/aws_cargohold-asset-storage/outputs.tf
```diff
@@ -0,0 +1,18 @@
...
+}
+
+output "cargohold_access_key" {
+  value = aws_iam_access_key.cargohold.id
+}
+
+# This value is sensitive in nature and you cannot get it from AWS later.
+output "cargohold_access_secret" {
+  value = aws_iam_access_key.cargohold.secret
+}
```
jschaul, line 18: As I commented on the previous PR on the other repo, this is not sustainable (see inline comment around the same code there).

lucendio:
> The cloudfront setup is missing. Are you planning to do this in this PR or in a subsequent PR? 

What are you referring to? The modules are solely based on @tiago-loureiro 's https://github.com/wireapp/wire-server-engagements/pull/10

terraform/modules/aws_cargohold-asset-storage/README.md
lucendio: Fair point. Consistency is a good thing. I now ended up using dashes in folders and underscores for any TF code. Would this be feasible?

terraform/modules/aws_cargohold-asset-storage/variables.tf
lucendio: it is indeed used, but only or generating the output `s3_endpoint`. Though, that specific output is a bit useless IMHO, since it is fairly static.

terraform/modules/aws_cargohold-asset-storage/outputs.tf
lucendio: Since it is a TF output, it is stored in the `.tfstate` and can be queried at all times with `terraform output`. So the comment is correct but also misleading. I adjusted it. But I think it is not faesible to put the comment on all outputs in any module which expose sensitive data

terraform/modules/aws-cargohold-asset-storage/README.md
```diff
@@ -0,0 +1,17 @@
+Terraform module: Cargohold Asset Storage
+=========================================
+
+This module creates an Object Storage on AWS for cargohold to store encrypted assets.
+
+
+#### How to use the module
+
+```hcl-terraform
```
jschaul, line 9: `hcl-terraform` is not a valid syntax. Use either `terraform` or `hcl`, which [should be supported](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml), but `hcl-terraform` is not.

terraform/modules/aws-cargohold-asset-storage/variables.tf
```diff
@@ -0,0 +1,17 @@
+variable "region" {
```
jschaul, line 1: This is misleading. While the output outputs a static string (which is not really useful) containing this region, nothing in this module actually ensures that the bucket is stored in that region. So if the user of the module configures their provider to use another region, this will be wrong. 

A few options I see:

1. Remove this variable and output
2. Use the region for the bucket creation in "aws_s3_bucket". There is a `region =` argument.

In all cases, the output should use the region from the "aws_s3_bucket" attributes, not from the input variable (as that could change, and then be wrong). See https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#attributes-reference

terraform/modules/aws-cargohold-asset-storage/README.md
```diff
@@ -0,0 +1,17 @@
+Terraform module: Cargohold Asset Storage
+=========================================
+
+This module creates an Object Storage on AWS for cargohold to store encrypted assets.
```
jschaul, line 4: This module is experimental, and lacks cloudfront setup. Can you add this to the readme? (see also the other readmes)

terraform/modules/aws-cargohold-asset-storage/variables.tf
lucendio: thx very much for pointing this out!

terraform/modules/aws_cargohold-asset-storage/outputs.tf
lucendio: never mind. saw [your comment](https://github.com/wireapp/wire-server-deploy/pull/244#discussion_r416607447)

PR #244, opened 2020-04-28
Add two TF modules for brig

lucendio:
prekey storage and queue for internal events are required by brig, whereasthe services for emailing is not needed and could be provided in another way (see chart value `brig.config.useSES`)


terraform/modules/aws-brig-email-sending/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+
+#### How to use the module
+
+```hcl-terraform
+module "bring_email_sending" {
+  source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-brig-email-sending?ref=develop"
+  
+  environment = "staging"
+  sender_address = "no-reploy@wire.com"
```
jschaul, line 14:
```suggestion
  sender_address = "no-reply@example.com"
```

terraform/modules/aws-brig-email-sending/README.md
```diff
@@ -0,0 +1,18 @@
+Terraform module: Brig to enable email sending
+==============================================
+
+This module creates AWS resources (SES, SQS, SNS) enabling Brig to send emails.
```
jschaul, line 4:
```suggestion
Wire-server's "brig" components needs to send emails. This can either be done by configuring an SMTP server (Option 1), or by using AWS resources (SES, SQS, SNS) (Option 2).

This terraform module enables brig to send emails using option 2.
```

terraform/modules/aws-brig-email-sending/outputs.tf
```diff
@@ -0,0 +1,14 @@
...
+
+output "ses_endpoint" {
+  value = "https://email.${var.region}.amazonaws.com"
+}
+
+output "brig_access_key" {
+  value = aws_iam_access_key.brig.id
+}
+
+# This value is sensitive in nature and you cannot get it from AWS later.
```
jschaul, line 11: Please apply the same refactoring on this PR than on the other PR about this comment.

terraform/modules/aws-brig-email-sending/outputs.tf
```diff
@@ -0,0 +1,14 @@
+# Output required to configure wire-server
+
+output "ses_endpoint" {
+  value = "https://email.${var.region}.amazonaws.com"
```
jschaul, line 4: Don't rely on input variables; instead, use the variables obtained from one of the resources that contain a region attribute.

terraform/modules/aws-brig-email-sending/README.md
```diff
@@ -0,0 +1,18 @@
...
+
```
jschaul, line 5: This module stores sensitive secrets inside terraform state file. Please add a comment in the readme warning people about this (i.e. ensure to encrypt the terraform state).

This module is also untested, and while we can merge an initial version to test the functionality and improve on subsequent PRs, please indicate that this code is not ready yet in the readme. See also the readmes of the other terraform modules for this.

terraform/modules/aws-brig-prekey-storage-event-queue/README.md
```diff
@@ -0,0 +1,18 @@
+Terraform module: Brig pre-key storage & event queue 
+===================================================
+
+This module creates AWS resources (SQS, DynamoDB) that allows Brig to store pre-keys on AWS and establishes a message
```
jschaul, line 4: Prekeys are not stored on DynamoDB. DynamoDB is purely used as a distributed locking mechanism to avoid race conditions on prekey insertion and retrieval. All storage happens in cassandra. Please change the module name, the header in this file, and the line above to reflect that (and remove the word "storage").

terraform/modules/aws-brig-email-sending/outputs.tf
lucendio: I guess you are referring to [this](https://github.com/wireapp/wire-server-deploy/pull/243/files#diff-37fe5bb4975cf90f7786a6ad418fab3dR15-R17) change. Since you didnt reply [here](https://github.com/wireapp/wire-server-deploy/pull/243#discussion_r416514296), I guess, this was what you had in mind?

terraform/modules/aws-brig-email-sending/outputs.tf
lucendio: No resource in the module exposes this information - as far as I can tell. So instead, I added ad provider directive and set its region with the given variable. and then, I use the `aws_region.current.name` from tiago's code for the interpolation. I tested this, and it works. This way, one can control in which region all resources of the module are going to be created.

terraform/modules/aws-brig-prekey-storage-event-queue/README.md
lucendio: done. please confirm whether this is suitable now.

terraform/modules/aws-brig-email-sending/outputs.tf
lucendio: never mind. saw [your comment](https://github.com/wireapp/wire-server-deploy/pull/244#discussion_r416607447)

terraform/modules/aws-brig-prekey-lock-event-queue/README.md
```diff
@@ -0,0 +1,28 @@
+Terraform module: Brig pre-key locking & event queue 
+====================================================
+
+State: __experimental__
+
+This module allows Brig to leverage AWS resources (A) to lock pre-keys during insertion and
+retrieval, and (B) to establish a message queue for internal events.
```
jschaul, line 7:
```suggestion
This module allows wire-server's brig service to leverage AWS resources (A) to acquire a lock using dynamoDB (used during insertion and retrieval of prekeys in cassandra to avoid race conditions), and (B) to establish a message queue for internal events (used e.g. during user deletions).
```

terraform/modules/aws-brig-email-sending/resources.dns.tf
```diff
@@ -0,0 +1,16 @@
+resource "aws_route53_record" "domain_amazonses_verification_record" {
```
jschaul, line 1: This file could use some comments to explain what it does. Alternatively, the REAMDE should explain the DKIM setup (or link to documentation on this elsewhere)

terraform/modules/aws-brig-email-sending/README.md
```diff
@@ -0,0 +1,33 @@
...
+mandatory.
+
+
+#### How to use the module
+
+```hcl
+module "brig_email_sending" {
+  source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-brig-email-sending?ref=develop"
+  
+  environment = "staging"
```
tiago-loureiro, line 27:
```suggestion
  environment = "dev"
```

Really nit-picky 🙈

terraform/modules/aws-brig-email-sending/README.md
```diff
@@ -0,0 +1,33 @@
+Terraform module: Brig to enable email sending
+==============================================
+
+State: __experimental__
+
+Wire-server's "brig" components needs to send emails. This can either be done by configuring an
```
tiago-loureiro, line 6:
```suggestion
Wire-server's "brig" components needs to send emails and scheduling user deletions. This can either be done by configuring an
```

tiago-loureiro: Nevermind my comment

terraform/modules/aws-brig-prekey-lock-event-queue/resources.dynamodb.tf
```diff
@@ -0,0 +1,15 @@
+# FUTUREWORK: Potentially look at autoscaling for dynamoDB
+# see: https://www.terraform.io/docs/providers/aws/r/appautoscaling_policy.html
+#
+resource "aws_dynamodb_table" "prekey_locks" {
+  name           = "${var.environment}-brig-prekay-locks"
```
tiago-loureiro, line 5:
```suggestion
  name           = "${var.environment}-brig-prekey-locks"
```

terraform/modules/aws-brig-prekey-lock-event-queue/README.md
```diff
@@ -0,0 +1,30 @@
+Terraform module: Brig pre-key locking & event queue
```
tiago-loureiro, line 1: We need a single user, scoped to those permissions allowed on AWS.

terraform/modules/aws-brig-email-sending/README.md
lucendio: it defaults to `dev` actually, that is why I decided to give it a different value, otherwise it could be omitted altogether. But then, the example doesnt really makes sense anymore.

terraform/modules/aws-brig-prekey-lock-event-queue/README.md
lucendio: done with the last refactoring (putting the two modules back together). pls have another look.

lucendio:
The following error
```
Error: Invalid count argument

  on ../../../../../wire-server-deploy/terraform/modules/aws-brig-prekey-lock-event-queue-email-sending/resources.dns.mailing.tf line 17, in resource "aws_route53_record" "ses_domain_dkim_record":
  17:   count   = var.enable_email_sending ? length( aws_ses_domain_dkim.brig[0].dkim_tokens ) : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.
```

has been worked around by setting the count of `ses_domain_dkim_record` to `count   = var.enable_email_sending ? 3 : 0` instead of dynamically obtaining the length of the token list.

`terraform plan` now works for `enable_email_sending = false` as well as `enable_email_sending = true`

terraform/modules/aws-brig-prekey-lock-event-queue-email-sending/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+##### With email sending __disabled__
+
+```hcl
+module "brig_prekey_lock_and_event_queue" {
+  source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-brig-prekey-lock-event-queue-email-sending?ref=develop"
+  
+  environment = "staging"
+  install_mailing = false    # default: true
```
tiago-loureiro, line 48:
```suggestion
  enable_email_sending = false    # default: true
```

terraform/modules/aws-brig-prekey-lock-event-queue-email-sending/resources.dynamodb.tf
```diff
@@ -0,0 +1,15 @@
+# FUTUREWORK: Potentially look at autoscaling for dynamoDB
```
tiago-loureiro, line 1: Right, I had a look at it: https://github.com/zinfra/backend-issues/issues/1319 but that was done with ansible. Once it makes its way to `prod`, I think we can adjust/adapt this.

terraform/modules/aws-brig-prekey-lock-event-queue-email-sending/resources.iam.tf
```diff
@@ -0,0 +1,57 @@
+resource "aws_iam_user" "brig" {
+  name          = "${var.environment}-brig-services"
+  force_destroy = true
```
tiago-loureiro, line 3: We called the user `<env>-<service>-full-access` [here](https://github.com/wireapp/wire-server-deploy/blob/develop/terraform/modules/aws-gundeck-push-notifications/resources.iam.tf#L2). Can we keep it like that?

PR #249, opened 2020-05-03
added possibility to specify proxy env vars in Ansible inventory

orandev:


ansible/admin_users.yml
```diff
@@ -5,6 +5,7 @@
 # See the `-k` and `-K` flags when running ansible-playbook.
 #
 - hosts: all
+  environment: "{{ proxy_env }}"
```
jschaul, line 8: Hi, could you change all these lines to

```suggestion
  environment: "{{ proxy_env | default({}) }}"
```
? This way, the playbook can be run as before for those people who do not need any proxy environment variables without any changes.

Otherwise, this looks good, thanks!

Issue #264, opened 2020-05-14
getting external smtp to work - smtpUsername to username?


tofutim:
Hi, I have been having a hard time getting email working on the demo server (https://github.com/wireapp/wire-docs/issues/56). In particular, I am attempting to use authenticated smtp. In `wire-server-deploy/charts/brig/templates/configmap.yaml`, I see
```
        smtpCredentials:
          username: {{ .smtp.username }}
          password: {{ .smtp.passwordFile }}
```
yet in wire-server/services/brig/src/Brig/Options.hs
```
data EmailSMTPCredentials = EmailSMTPCredentials
  { -- | Username to authenticate
    --   against the SMTP server
    smtpUsername :: !Text,
    -- | File containing password to
    --   authenticate against the SMTP server
    smtpPassword :: !FilePathSecrets
  }
  deriving (Show, Generic)

instance FromJSON EmailSMTPCredentials
```
it gets read in App.hs
```
      smtpCredentials <- case Opt.smtpCredentials s of
        Just (Opt.EmailSMTPCredentials u p) -> do
          pass <- initCredentials p
          return $ Just (SMTP.Username u, SMTP.Password pass)
        _ -> return Nothing
```
where I think there will be a mismatch between smtpUsername and username, and smtpPassword and password, maybe why I have the error
```
service: user error (AesonException "Error in $.emailSMS.email.smtpCredentials: parsing Brig.Options.EmailSMTPCredentials(EmailSMTPCredentials) failed, key \"smtpUsername\" not found")
```
for my deploy values.txt, I am now using
```
    smtp:
      host: smtp.gmail.com # change this if you want to use your own SMTP server
      port: 587        # change this
      connType: tls # change this. Possible values: plain|ssl|tls
      username: xxxxx@gmail.com
      passwordFile: /etc/wire/brig/secrets/smtp-password.txt
```

I'd like to test out my theory, but haven't yet explored compiling brig on this end. Any thoughts much appreciated. I think perhaps I can make the changes in configmap.yaml and
```
./bin/update.sh ./charts/<chart-name> # this will clean and re-package subcharts
helm install charts/<chart-name> # specify a local file path
```

**Update.** I tried changing the configmap and then
```
helm dependency update charts
```
from /mnt/wire-server-deploy, then
```
helm upgrade --install wire-server charts/wire-server -f wire-server/values.yaml -f wire-server/secrets.yaml
```
this time, a new error
```
[...]

PR #265, opened 2020-05-14
fix smtpCredentials to match EmailSMTPCredentials in brig Options.hs

tofutim:
to fix issue #264 - though I think it would be prettier to change Option.hs instead so that you have 
```
data EmailSMTPCredentials = EmailSMTPCredentials
  { -- | Username to authenticate
    --   against the SMTP server
    username :: !Text,
    -- | File containing password to
    --   authenticate against the SMTP server
    password :: !FilePathSecrets
  }
  deriving (Show, Generic)
```
but this will get things working without having to monkey around with wire-server

### DIFF
--- a/charts/brig/templates/configmap.yaml
+++ b/charts/brig/templates/configmap.yaml
@@ -64,8 +64,8 @@ data:
         smtpConnType: {{ .smtp.connType }}
         {{- if .smtp.username }}
         smtpCredentials:
-          username: {{ .smtp.username }}
-          password: {{ .smtp.passwordFile }}
+          smtpUsername: {{ .smtp.username }}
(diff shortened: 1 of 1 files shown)


lucendio:
@tiago-loureiro I'm able to retrace @tofutim 's reasoning and I looked at the history of ` Options.hs`. It appears that those fields where never called `username` & `password`. Do you consent? If so, please approve, since they are your LOCs.

> to fix issue #264 - though I think it would be prettier to change Options.hs instead so that you have

Looking at the impact of dealing with it this way, I'd say, we fix the automation part instead.

lucendio:
Fixes https://github.com/wireapp/wire-server-deploy/issues/264

Issue #266, opened 2020-05-14
Emails not received after register


rahulit1991:
Hello,
i am facing an issue after completely setup demo on virtual server. everything installed properly. but after register user i am not able to receive email in my mailbox. while debug logs of smtp pod, i see this errors.can you please help to resolve this error?

**kubectl log -f demo-smtp-84b7b85ff6-p62q6**

Size of off_t: 8
  289 delivering 1jZD6i-00004e-Oj
  289 R: dnslookup for rahul******@hotmail.com
  290 T: remote_smtp for rahul*****@hotmail.com
  288 LOG: smtp_connection MAIN
  288   SMTP connection from 10-233-64-34.brig.default.svc.cluster.local (brig-bf8d7fcd5-5pbsx) [10.233.64.34] lost D=5s
  287 Connecting to hotmail-com.olc.protection.outlook.com [104.47.14.33]:25 ... failed: Connection timed out (timeout=5m)
  287 LOG: MAIN
  287   H=hotmail-com.olc.protection.outlook.com [104.47.14.33] Connection timed out
  290 Connecting to hotmail-com.olc.protection.outlook.com [104.47.59.161]:25 ... failed: Connection timed out (timeout=5m)
  290 LOG: MAIN
  290   H=hotmail-com.olc.protection.outlook.com [104.47.59.161] Connection timed out

appreciate for your help.

lucendio:
Hi @rahulit1991,

The logs reading:

> 287 Connecting to hotmail-com.olc.protection.outlook.com [104.47.14.33]:25 ... failed: Connection timed out (timeout=5m)

Can you verify whether that server is reachable from within the cluster?

rahulit1991:
yes. i setup demo on AWS EC2 instance. Also perform curl -i https://nginz-https.<domain>/status
. and its working well. All security ports are open. If you can help me to how to debug this issue differently, then you are appreciate. i perform all things that mention in installation documents.https://docs.wire.com/how-to/install/helm.html

lucendio:
So, I assume that you adjusted [values](https://github.com/wireapp/wire-server-deploy/blob/develop/values/wire-server/demo-values.example.yaml#L47) and [secrets](https://github.com/wireapp/wire-server-deploy/blob/develop/values/wire-server/demo-secrets.example.yaml#L3) according to your setup? Could you pelase those files?

rahulit1991:
[...]

PR #268, opened 2020-05-18
Make field sesQueue to be required in case AWS SES is being configured

lucendio:
* helm doesn't seem to fail if a sub-field is not set, not even if it's not
  defined as default in $chartName/values.yaml
* case: `aws.useSES: true` and `./values` example files don't list it as
* ths commit introduced the expected behaviour: throw if required while not
  mentioned in any of the example values files

Fixes https://github.com/wireapp/wire-server-deploy/issues/266

### DIFF
--- a/charts/brig/templates/configmap.yaml
+++ b/charts/brig/templates/configmap.yaml
@@ -55,7 +55,7 @@ data:
     emailSMS:
       email:
       {{- if .useSES }}
-        sesQueue: {{ .aws.sesQueue }}
+        sesQueue: {{ required "Missing value: brig.aws.sesQueue" .aws.sesQueue }}
         sesEndpoint: {{ .aws.sesEndpoint | quote }}
       {{- else }}
         smtpEndpoint:


charts/brig/templates/configmap.yaml
```diff
@@ -55,7 +55,7 @@ data:
     emailSMS:
       email:
       {{- if .useSES }}
-        sesQueue: {{ .aws.sesQueue }}
+        sesQueue: {{ required "Missing value: brig.aws.sesQueue" .aws.sesQueue }}
         sesEndpoint: {{ .aws.sesEndpoint | quote }}
```
akshaymankar, line 59: We should also put this check for `sesEndpoint`, brig will fail with the same confusing error if `sesEndpoint` isn't defined.

akshaymankar: Hmm, then I guess it is ok. If someone sets it to `null`, it is on them :smile:

PR #272, opened 2020-05-21
[TF modules] Ensure uniqueness of cargohold bucket name

lucendio:
since buckets are AWS-global (or at least region-global) the
naming so far was not unique enough, so we use a random suffix,
which will change every time env or name changes (see, `keepers`)

### DIFF
--- a/terraform/modules/aws-cargohold-asset-storage/resources.s3.tf
+++ b/terraform/modules/aws-cargohold-asset-storage/resources.s3.tf
@@ -11,7 +11,7 @@ resource "random_string" "bucket" {
   number = true
   special = false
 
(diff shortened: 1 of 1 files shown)


terraform/modules/aws-cargohold-asset-storage/resources.s3.tf
```diff
@@ -1,5 +1,18 @@
...
 }
+
+resource "random_string" "bucket" {
+  length = 8
+  lower = true
+  upper = false
+  number = true
+  special = false
+
+  keepers {
```
lucendio, line 14:
```suggestion
  keepers = {
```

PR #276, opened 2020-05-25
Comment on email visibility

tiago-loureiro:


akshaymankar:
Isn't [wire-server/docs/reference/config-options.md](https://github.com/wireapp/wire-server/blob/develop/docs/reference/config-options.md) a better place for this kind of documentation?

charts/brig/values.yaml
```diff
@@ -63,6 +63,11 @@ config:
     setDefaultLocale: en
     setMaxTeamSize: 500
     setMaxConvSize: 500
+    # Who can see the emails of other users
+    # visible_to_self: only the self user can
+    # visible_if_on_same_team: users in the same team can view each others' emails
+    # visible_if_on_team: any team user can have its email viewed
+    # Ref.: https://github.com/wireapp/wire-server/blob/0126651a25aabc0c5589edc2b1988bb06550a03a/services/brig/src/Brig/Options.hs#L304-L306
```
lucendio, line 70:
```suggestion
    # For meaning, see https://github.com/wireapp/wire-server/blob/0126651a25aabc0c5589edc2b1988bb06550a03a/services/brig/src/Brig/Options.hs#L304-L306
```
Referencing the wire-server code and therefore point to the meaning of the values totally make sense, but adding them here too is redundant and should be avoided IMHO. In case of change, nobody might be aware of the duplication here.

tiago-loureiro: Hmmm, to avoid duplication maybe we can just point to https://github.com/wireapp/wire-server/blob/develop/docs/reference/config-options.md and consistently describe the different options there? (And keep just the defaults here)

lucendio: pointing directly to code would be my personal preference, because usually changes appear there first (otherwise git log ftw), and sometimes that's the only place, but sure. In this case, your suggestion has the same result.

jschaul:
@akshaymankar @lucendio should we merge this change as-is, make a change, or close this PR without further action?

lucendio:
I'll take the liberty to just merge it but also open up a PR on wire-server as @akshaymankar suggested, If I understand correctly. And, I think it's still fine here to just point to the code.

lucendio (approved):
Approving my own change. Well, it's Friday ¯\_(ツ)_/¯

Issue #285, opened 2020-06-13
SMS configuration issues with wire server deploy script


ramesh8830:
Hi,

I am trying to deploy wire server using below SMS configuration and getting below error when I set smsSender as number in values.yaml file.

```
    emailSMS:
      general:
        emailSender: mail@example.com # change this
        smsSender: "+18123456789" # change this if SMS support is desired
```

Getting below error for brig pod.

`service: user error (AesonException "Error in $.emailSMS.general.smsSender: parsing Text failed, expected String, but encountered Number")`

Please help me.

rahulit1991:
facing same issue with twilio. first i set number but it through an error above error. After that i allowed alphanumeric id in account. but still facing same issue.

mheinzel:
Did you make sure to put quotes around the number? Otherwise, YAML will parse it as a number, not a string.

For example above, it says `smsSender: "+18123456789"`, which works for me, but without the quotes `smsSender: +18123456789`, I get the same error you pasted above.

rahulit1991:
yes. i tried with both. did not work.

ramesh8830:
> Did you make sure to put quotes around the number? Otherwise, YAML will parse it as a number, not a string.
> 
> For example above, it says `smsSender: "+18123456789"`, which works for me, but without the quotes `smsSender: +18123456789`, I get the same error you pasted above.

I tried with double quotations and didn't work.

lucendio:
Where did you try the quotation? In Helm's values.yaml or in the resulting ConfigMap that is consumed by Brig?

ramesh8830:
I tried it on the values.yaml inside wire-server

lucendio:
Okay, that wont work, see https://github.com/wireapp/wire-server-deploy/blob/develop/charts/brig/templates/configmap.yaml#L77.

I suggest to add the quotes directly in the ConfigMap (`data.brig.yaml`). Dont forget to trigger the pods so they get recreated (w/o redeploying!),  so that they pick up the changes in the ConfigMap. If that works, we need to adjust the brig chart.

ramesh8830:
Thank you @lucendio . That worked by adding quotes in the ConfigMap file.

rahulit1991:
@lucendio , i take latest pull and now i can see new changes made in configMap.yml. " **smsSender: {{ .emailSMS.general.smsSender | quote }}"**.

[...]

PR #287, opened 2020-06-15
Quote smsSender

lucendio:
A mobile number will be parsed into an integer, which is not compatible with the type definition is Brig's Haskell code

Fixes https://github.com/wireapp/wire-server-deploy/issues/285


mheinzel:
Ha, I only had a quick look yesterday, but was starting to suspect that the quotes were going missing during templating. Thanks for fixing it!

PR #290, opened 2020-06-19
Add ingress rule to allow NLB to talk to k8s

lucendio:
The origin of the ports can be found [here](https://github.com/wireapp/wire-server-deploy/blob/develop/charts/nginx-ingress-controller/values.yaml#L88-L89).

### DIFF
--- a/terraform/modules/aws-vpc-security-groups/main.tf
+++ b/terraform/modules/aws-vpc-security-groups/main.tf
@@ -210,6 +210,17 @@ resource "aws_security_group" "k8s_node" {
     security_groups = ["${aws_security_group.talk_to_k8s.id}"]
   }
 
+  # incoming from NLB
+  ingress {
+    from_port       = 31772
+    to_port         = 31773
+    protocol        = "tcp"
+    security_groups = [
+      # NOTE: NLBs dont allow security groups to be be set on them, which is why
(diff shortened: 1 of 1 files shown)


terraform/modules/aws-vpc-security-groups/main.tf
```diff
@@ -210,6 +210,16 @@ resource "aws_security_group" "k8s_node" {
     security_groups = ["${aws_security_group.talk_to_k8s.id}"]
   }
 
+  # incoming from NLB
+  ingress {
+    from_port       = 31772
+    to_port         = 31773
+    protocol        = "tcp"
+    # NOTE: NLBs dont allow security groups to be be set on them, which is why
+    # we go with the CIDR for now, which is hard-coded adn evil and needs fixing
```
mheinzel, line 219:
```suggestion
    # we go with the CIDR for now, which is hard-coded and evil and needs fixing
```

lucendio: you are on a row here! thx :D

mheinzel:
(missing some knowledge here, so I don't think I can confidently review this except for fixing typos, sorry)

lucendio:
Covered in https://github.com/wireapp/wire-server-deploy/pull/297 and more. So closing this one.

PR #294, opened 2020-06-23
[chart:team-settings] Rollback to v2.10

lucendio:
Version 2.11-2.12 are broken, because stripe broke in a minor version release.
According to web-team, setting `FEATURE_ENABLE_PAYMENT="false"` should have done
the trick, but it didnt. So we fall back to v2.10.
Furthermore, this makes `FEATURE_ENABLE_PAYMENT` defaulting to `false`. Since we
consider Helm deployments to be on-prem where billing for teams is not a thing.

This also fixes our example files. Each frontend app has to refer to its own FQDN.

### DIFF
--- a/charts/team-settings/templates/deployment.yaml
+++ b/charts/team-settings/templates/deployment.yaml
@@ -40,6 +40,10 @@ spec:
             value: https://{{ .Values.config.externalUrls.backendRest }}
           - name: BACKEND_WS
             value: wss://{{ .Values.config.externalUrls.backendWebsocket }}
+            # NOTE defaults to 'true', but since we assume on-prem here, we default to 'false'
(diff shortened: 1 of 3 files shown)


akshaymankar:
Also, if it wasn't already considered, we should explicitly call out the choice we're making: break large teams to make payments work.

lucendio:
Quick note on @akshaymankar - we had a quick chat. An information in the release note for sure makes sense.

At the time of writing, 2.13 is almost released. So, I'm holding on for a bit longer (before end of day). In case we get the release in time, we would not need to roll back but roll forward.

lucendio:
Closed in favour of https://github.com/wireapp/wire-server-deploy/pull/300

PR #295, opened 2020-06-23
[helm] Adjust scripts that are used in automation

lucendio:
* removed certain lines since they are redundant when using the toolchain
* there are more files under bin/ referring to `helm`, but I could find an
  usage outside of this code base

### DIFF
--- a/bin/sync.sh
+++ b/bin/sync.sh
@@ -49,21 +49,13 @@ else
     charts=( $(find $CHART_DIR/ -maxdepth 1 -type d | sed -n "s=$CHART_DIR/\(.\+\)=\1 =p") )
 fi
 
-# install s3 plugin
+# install s3 plugin if not present
 # See https://github.com/hypnoglow/helm-s3/pull/56 for reason to use fork
 s3_plugin_version=$(helm plugin list | grep "^s3 " | awk '{print $2}' || true)
 if [[ $s3_plugin_version != "0.9.0" ]]; then
     echo "not version 0.9.0 from steven-sheehy fork, upgrading or installing plugin..."
     helm plugin remove s3 || true
     helm plugin install https://github.com/steven-sheehy/helm-s3.git --version v0.9.0
-else
-    # double check we have the right version of the s3 plugin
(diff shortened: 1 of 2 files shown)


lucendio:
I found another solution. So this PR became a small clean up.

> All the tooling needed to work with wire-server-deploy need to be available from within wire-server-deploy.

Wwhile I generally agree with your statement, I'd love to see some use cases for the files that have been touched outside of our automation system. Maybe that helps me to clear thing up.

* sync.sh
* update.sh
* set-version.sh

jschaul:
Hm, okay, the force push makes it hard to follow my previous comment as the previous code changes are not visible anymore. Discussion taken to another place. For the random reader: The initial PR introduced environment variables, and the existence of custom wrapper scripts only part of another repository that is not open-source.

bin/sync.sh
```diff
@@ -49,21 +49,13 @@ else
...
-# install s3 plugin
+# install s3 plugin if not present
 # See https://github.com/hypnoglow/helm-s3/pull/56 for reason to use fork
 s3_plugin_version=$(helm plugin list | grep "^s3 " | awk '{print $2}' || true)
 if [[ $s3_plugin_version != "0.9.0" ]]; then
     echo "not version 0.9.0 from steven-sheehy fork, upgrading or installing plugin..."
     helm plugin remove s3 || true
     helm plugin install https://github.com/steven-sheehy/helm-s3.git --version v0.9.0
-else
-    # double check we have the right version of the s3 plugin
```
jschaul, line 60: not sure what the motivation is of removing this. Does it cause any problems? The version info is not specific enough, see also comments in https://github.com/wireapp/wire-server-deploy/pull/112

lucendio: double-checking , to me, is redundant here, but I might lack of a specific use case where this is needed. Furthermore, it uses `$HOME`, which makes some assumption on sth that is not necessarily correct for Helm v3 (`HELM_HOME` couls a) change this, and b) was replaced by those `XDG_[...]` vars).

jschaul: okay, thanks. Well, we can remove this now and in case we get some problems with wrong versions, fix that later.

PR #298, opened 2020-06-24
[tf-module:dns-records] Add a module to create some DNS records

lucendio:
The types are currently limited. We might need to add support for TXT in the future, too.

### DIFF
--- a/terraform/modules/aws-dns-records/README.md
+++ b/terraform/modules/aws-dns-records/README.md
@@ -24,7 +24,7 @@ module "dns_records" {
 }
 ```
 
-If not further specified, it created entries for the following FQDNs:
+If not further specified, it creates entries for the following FQDNs:
 
 * `nginz-https.staging.example.com`
 * `nginz-ssl.staging.example.com`
@@ -32,3 +32,7 @@ If not further specified, it created entries for the following FQDNs:
 * `assets.staging.example.com`
 * `account.staging.example.com`
 * `teams.staging.example.com`
+
+These sub-domains represent the primary set of FQDNs used in a
+[`wire-server` installation](https://docs.wire.com/how-to/install/helm-prod.html#how-to-set-up-dns-records),
+to expose all frontend applications as well as necessary HTTP & websocket endpoints.
--- a/terraform/modules/aws-dns-records/README.md
+++ b/terraform/modules/aws-dns-records/README.md
@@ -26,8 +26,8 @@ module "dns_records" {
 
 If not further specified, it created entries for the following FQDNs:
 
-* `https.staging.example.com`
-* `ssl.staging.example.com`
+* `nginz-https.staging.example.com`
+* `nginz-ssl.staging.example.com`
 * `webapp.staging.example.com`
 * `assets.staging.example.com`
 * `account.staging.example.com`
--- a/terraform/modules/aws-dns-records/README.md
+++ b/terraform/modules/aws-dns-records/README.md
@@ -1,0 +1,34 @@
+Terraform module: DNS records
+=============================
+
+State: __experimental__
+
+This module creates a set of DNS entries on AWS. As of now it can the following type of records:
+
+* A (`ips`)
+* CNAME (`cnames`)
+
+AWS resources: route53
+
+
+#### How to use the module
+
+```hcl
+module "dns_records" {
+  source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-dns-records?ref=develop"
+  
+  environment = "staging"
+
+  zone_fqdn = "example.com"
+  ips = [ "9.9.9.10", "23.42.23.42" ]
+}
+```
+
+If not further specified, it created entries for the following FQDNs:
+
+* `https.staging.example.com`
+* `ssl.staging.example.com`
+* `webapp.staging.example.com`
+* `assets.staging.example.com`
+* `account.staging.example.com`
+* `teams.staging.example.com`
--- a/terraform/modules/aws-dns-records/README.md
+++ b/terraform/modules/aws-dns-records/README.md
@@ -3,7 +3,7 @@ Terraform module: DNS records
 
 State: __experimental__
 
-This module creates a set of DNS entries on AWS. As of now it can the following type of records:
+This module creates a set of DNS entries on AWS. As of now it's capable of managing the following type of records:
 
 * A (`ips`)
 * CNAME (`cnames`)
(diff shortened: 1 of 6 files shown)


terraform/modules/aws-dns-records/README.md
```diff
@@ -0,0 +1,34 @@
...
+#### How to use the module
+
+```hcl
+module "cargohold_asset_storage" {
+  source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-dns-records?ref=develop"
+  
+  environment = "staging"
+
+  zone_fqdn = "example.com"
+  ips = [ "256.256.256.256" ]
```
julialongtin, line 23: 256? don't you mean 255?

terraform/modules/aws-dns-records/README.md
```diff
@@ -0,0 +1,34 @@
...
+}
+```
+
+If not further specified, it created entries for the following FQDNs:
+
+* `https.staging.example.com`
+* `https.staging.example.com`
```
julialongtin, line 30: ssl?

terraform/modules/aws-dns-records/README.md
lucendio: I wanted to use sth that does not exist, but apparently it's more confusing than it helps.

terraform/modules/aws-dns-records/README.md
lucendio: copy&paste mistake right there

terraform/modules/aws-dns-records/variables.tf
```diff
@@ -0,0 +1,53 @@
...
+  description = "name of the sub-tree all given subdomains are append to (defaults to $environment; example: $subdomains[0].$domain.$zone_fqdn)"
+  default     = null
+}
+
+variable "subdomains" {
+  type        = list(string)
+  description = "list of sub-domains that will be registered under the given root domain"
+  default = [
+    "https",
+    "ssl",
```
jschaul, line 23: Please adhere to the names in https://docs.wire.com/how-to/install/helm.html#how-to-set-up-dns-records

terraform/modules/aws-dns-records/README.md
```diff
@@ -0,0 +1,34 @@
+Terraform module: DNS records
+=============================
+
+State: __experimental__
+
+This module creates a set of DNS entries on AWS. As of now it can the following type of records:
```
jschaul, line 6: DNS entries for what purpose? What should those IPs be? Why is it useful?

there are a few assumtions here that would be nice to make explicit in the readme (please reformulate the following):

* one or multiple kubernetes servers with a public IP or load balancers fronting kubernetes servers, on which wire-server's nginx-ingress listens on port 443
* basic round-robin DNS? What happens when one machine goes down?
* this is about wire-server DNS as described on docs.wire.com / find-the-right-page.
* this does not create records for restund servers

terraform/modules/aws-dns-records/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+  zone_fqdn = "example.com"
+  ips = [ "9.9.9.10", "23.42.23.42" ]
+}
+```
+
+If not further specified, it created entries for the following FQDNs:
+
+* `https.staging.example.com`
+* `ssl.staging.example.com`
```
jschaul, line 30: please use the names from the docs; or, if you want different names, we need a discussion and then change those names everywhere in all repositories. For the moment, let's please be consistent.

jschaul:
Renaming the module so that it's clearer what it does would be nice. "DNS records" is too generic.

terraform/modules/aws-dns-records/variables.tf
lucendio: Would you be open to rather adjust the docs, since the docs also refer to the use of https://github.com/wireapp/wire-server-deploy/blob/develop/ansible/roles/minio-static-files/defaults/main.yml#L9-L14, where the module actually adheres.

terraform/modules/aws-dns-records/README.md
lucendio: > there are a few assumptions here that would be nice to make explicit in the readme (please reformulate the following):

I probably did't quite understand your comment. So, can you point me to the places where this module makes any of the assumptions that you mentioned (other than the default value for `subdomains`)?

lucendio:
> Renaming the module so that it's clearer what it does would be nice. "DNS records" is too generic.

What would you have in mind?

terraform/modules/aws-dns-records/variables.tf
jschaul: https://github.com/wireapp/wire-server-deploy/blob/develop/ansible/roles/minio-static-files/defaults/main.yml#L9-L14 is incorrect and should be changed.

Currently the docs, and all the DNS names used there, as well as in all the demo/prod/default values should be used for the purpose of this PR.

If you want to change names, please 1) open an issue with a motivation, and 2) make a PR that ensures consistency here across all DNS names in all of our configuration files - but make this a separate PR, not a side effect of a PR introducing a new terraform module. Thanks.

terraform/modules/aws-dns-records/README.md
jschaul: I suppose the purpose / intention / reason for existence of the module (as seen by its default subdomains list) is as a way to create DNS records pointing to a wire-server installation. I think we don't gain much on making this module very generic, I think it's most useful when geared towards the creation of the DNS records for a wire-server installation. So specifying this in the README, to me, is useful. Would you be fine adding a sentence or two about this to the README?

It's true that in theory someone could also take this module and use it for another purpose by overriding the subdomains variable. But for that, someone needs to actually look at the source code of the module to understand that. It would be nice if the purpose and usage of this terraform module was clearer by only looking at the README.

lucendio: > actually look at the source code of the module to understand that

Even though I might not given the impression, but I'm a huge fan of writing readable & self-explanatory (and hopefully in that way more maintainable) code. Therefore, I encourage anyone who wants to use that code, also to have a quickly look into it, before using it. In this particular case, it's even not that much code. Also, since this is not compiled code, and actually used where its located, it's very accessible (copy in `.terraform`).

Why this kind of thinking? The past has taught me: *Docs are not to be trusted (outdated, flawed,misleading). Look at the code if you want to know what's going on.*

That being said, It's just my personal point of view. If we, as a team, follow a different mantra, I'd be happy to adjust.
Anyway, you have a point here. I will add a note about as to why this set of default subdomains exactly.

> It's true that in theory 

For starters, I will also use it to create records for SFT machines. With this, the *theory* becomes *practice* :)

PR #305, opened 2020-06-30
add a bash script for talking to S3 with AWS authentication V4. for t…

julialongtin:
…esting S3 connection during installation


terraform/modules/aws-vpc-security-groups/main.tf
```diff
@@ -191,11 +191,34 @@ resource "aws_security_group" "talk_to_k8s" {
     cidr_blocks = ["172.17.0.0/20"]
   }
 
+  # the application itsself.
+  egress {
+    from_port   = 31772
+    to_port     = 31773
+    protocol    = "tcp"
+    cidr_blocks = ["172.17.0.0/20"]
+  }
```
lucendio, line 200: As much I would love to see this change going through, I think it's more related to https://github.com/wireapp/wire-server-deploy/pull/299, rather than this PR - or maybe adjust the commit message...

julialongtin: the TF is not supposed to be here. i'll remove it, for a seperate PR.

PR #310, opened 2020-07-09
remove frontend helm charts, another way of hosting these needs to be…

jschaul:
… found


jschaul:
It would be great if we could schedule a discussion on the topic at some point. But yes, this was just a way to provoke a discussion. It didn't work so far (discussion didn't happen so far).

PR #311, opened 2020-07-10
S3 support

julialongtin:


terraform/modules/aws-cargohold-asset-storage/resources.s3.tf
```diff
@@ -24,3 +24,29 @@ resource "random_string" "bucket" {
...
+
+  filter {
+    name   = "association.subnet-id"
+    values = var.subnet_ids
+  }
+}
+
+# the routing table association that allows nodes to route traffic to the S3 endpoint.
+resource "aws_vpc_endpoint_route_table_association" "private_s3" {
+  count = length(data.aws_route_tables.private.ids)
```
lucendio, line 48: I think we should adjust this so that it wont break when the order changes, see: https://github.com/wireapp/wire-server-deploy/pull/299/files#diff-6c639265bb28ba18ca37d47a7832090fR37

terraform/modules/aws-cargohold-asset-storage/resources.s3.tf
```diff
@@ -24,3 +24,29 @@ resource "random_string" "bucket" {
...
+
+  route_table_id  = tolist(data.aws_route_tables.private.ids)[count.index]
+  vpc_endpoint_id = aws_vpc_endpoint.s3.id
+}
```
lucendio, line 52: > I think we should adjust this so that it wont break when the order changes, see: https://github.com/wireapp/wire-server-deploy/pull/299/files#diff-6c639265bb28ba18ca37d47a7832090fR37

```suggestion
resource "aws_vpc_endpoint_route_table_association" "private_s3" {
  for_each = { for k, v in data.aws_route_tables.private.ids : v => v }

  route_table_id  = each.value
  vpc_endpoint_id = aws_vpc_endpoint.s3.id
}
```
Please give it another spin, in case you are going to integrate it.

PR #314, opened 2020-07-14
Provide AWS_REGION variable to cargohold

mheinzel:
See https://github.com/zinfra/backend-issues/issues/488

Now with [cargohold using V4 signatures](https://github.com/wireapp/wire-server/pull/1157), the region is part of the Authorization header and needs to configured correctly. We already do that for the other services, so I just copied the code from there.

### DIFF
--- a/charts/cargohold/templates/tests/cargohold-integration.yaml
+++ b/charts/cargohold/templates/tests/cargohold-integration.yaml
@@ -17,4 +17,12 @@ spec:
     volumeMounts:
     - name: "cargohold-integration"
       mountPath: "/etc/wire/integration"
+    env:
+    # these dummy values are necessary for Amazonka's "Discover"
+    - name: AWS_ACCESS_KEY_ID
+      value: "dummy"
+    - name: AWS_SECRET_ACCESS_KEY
+      value: "dummy"
+    - name: AWS_REGION
+      value: "eu-west-1"
   restartPolicy: Never
--- a/charts/cargohold/templates/deployment.yaml
+++ b/charts/cargohold/templates/deployment.yaml
@@ -54,6 +54,8 @@ spec:
               secretKeyRef:
                 name: cargohold
                 key: awsSecretKey
+          - name: AWS_REGION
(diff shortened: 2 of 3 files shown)


charts/cargohold/templates/tests/cargohold-integration.yaml
```diff
@@ -17,4 +17,12 @@ spec:
...
     - name: "cargohold-integration"
       mountPath: "/etc/wire/integration"
+    env:
+    # these dummy values are necessary for Amazonka's "Discover"
+    - name: AWS_ACCESS_KEY_ID
+      value: "dummy"
+    - name: AWS_SECRET_ACCESS_KEY
+      value: "dummy"
+    - name: AWS_REGION
+      value: "eu-west-1"
```
mheinzel, line 27: This is what we also do for the other services, but unless I'm missing something it is completely unnecessary?
The integration test binaries don't talk to AWS directly, so they should not need these variables.

We should probably remove them (from all integration pods, in a separate PR).

tiago-loureiro: I cannot remember right now but... amazonka might complain/throw errors if they are not defined?

mheinzel: But this pod only runs the `amazonka-integration` binary, right? And that doesn't depend on amazonka.

mheinzel: I also verified this locally for gundeck (from where I copied the config). I can run the binary without providing the environment variables.

Well, for now I'll merge this and then I can make a separate PR removing the config from all `-integration` pods.

tiago-loureiro: `cargohold-integration` depends on `cargohold` lib which does have a dependency on `amazonka`. I can't tell for sure if it's a problem or not; it may be different for other services because `brig-integration` and `galley-integration` do talk directly to AWS (they check the queues)

PR #315, opened 2020-07-14
[tf-module:dns-records] Add output for FQDNs

lucendio:
Since the FQDNs are assembled within the module, it make sense to expose those, so that the same logic is not getting re-implemented.

### DIFF
--- a/terraform/modules/aws-dns-records/outputs.tf
+++ b/terraform/modules/aws-dns-records/outputs.tf
@@ -1,0 +1,6 @@
+output "fqdns" {
+  value = concat(
+    [for record in aws_route53_record.a : record.fqdn],
+    [for record in aws_route53_record.cname : record.fqdn]
+  )
+}


terraform/modules/aws-dns-records/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "fqdns" {
+  value = [for v in var.subdomains: join(".", concat([v], local.name_suffix))]
```
flokli, line 2: This replicates the same logic used in https://github.com/wireapp/wire-server-deploy/blob/develop/terraform/modules/aws-dns-records/resources.route53.tf#L5 and https://github.com/wireapp/wire-server-deploy/blob/develop/terraform/modules/aws-dns-records/resources.route53.tf#L16.

Wouldn't it make more sense to move logic from these places into a variable, and export these variable values (and/or a combined list of it) as output?

lucendio: Since the function call in those two lines refer to diferent context on each iteration, moving it into a `local` (if that's what you mean by *variable*) approach wouldn't really work iMHO. But concatenating the name from each resource might work. Let's see if TF handles `for_each = []` as expected ...

lucendio: What about now?

flokli: Neat!

PR #322, opened 2020-07-31
Provide a way to configure SFT service discovery

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1453

### DIFF
--- a/charts/brig/templates/configmap.yaml
+++ b/charts/brig/templates/configmap.yaml
@@ -150,6 +150,19 @@ data:
       configTTL: 3600 # 1 hour
       tokenTTL: 43200 # 12 hours
 
+    {{- if .sft }}
+    {{- with .sft }}
+    sft:
+      sftBaseDomain: {{ required "Missing value: .sft.sftBaseDomain" .sftBaseDomain }}
+      {{- if .sftSRVServiceName }}
+      sftSRVServiceName: {{ .sftSRVServiceName }}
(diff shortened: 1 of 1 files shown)


charts/brig/templates/configmap.yaml
```diff
@@ -150,6 +150,19 @@ data:
...
 
+    {{- if .sft }}
+    {{- with .sft }}
+    sft:
+      sftBaseDomain: {{ required "Missing value: .sft.sftBaseDomain" .sftBaseDomain }}
+      {{- if .sftSRVServiceName }}
+      sftSRVServiceName: {{ .sftSRVServiceName }}
+      {{- end }}
+      {{- if .sftDiscoveryIntervalSeconds }}
+      sftSRVServiceName: {{ .sftDiscoveryIntervalSeconds }}
```
flokli, line 161: This is a duplicate key.

Also, I wonder why we assemble `sftBaseDomain` and `sftSRVServiceName` in the backend, and not just pass the fully qualified record name to the server.

fisx (approved):
Indentation is odd, but maybe there is a good reason for that.  Anyway looks sound to me.

akshaymankar:
> Indentation is odd, but maybe there is a good reason for that. Anyway looks sound to me.

The reason is that we are templating to output some indentation sensitive code (YAML) :cry:

PR #324, opened 2020-08-03
Update TLS config

franziskuskiefer:
Fixes https://github.com/zinfra/backend-issues/issues/1628

Note that this also disables TLS versions we don't want and enables TLS 1.3. Le me know when there's a reason why that doesn't work.

### DIFF
--- a/charts/backoffice/templates/configmap.yaml
+++ b/charts/backoffice/templates/configmap.yaml
@@ -126,8 +126,8 @@ data:
         ssl_session_cache builtin:1000 shared:SSL:10m;
         ssl_session_timeout 5m;
         ssl_prefer_server_ciphers on;
-        ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-        ssl_ciphers 'EECDH+ECDSA+AESGCM EECDH+aRSA+AESGCM EECDH+ECDSA+SHA384 EECDH+ECDSA+SHA256 EECDH+aRSA+SHA384 EECDH+aRSA+SHA256 EECDH+aRSA+RC4 EECDH EDH+aRSA RC4 !aNULL !eNULL !LOW !3DES !MD5 !EXP !PSK !SRP !DSS';
+        ssl_protocols TLSv1.2 TLSv1.3;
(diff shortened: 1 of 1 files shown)


franziskuskiefer:
> [`1.15.12`](https://github.com/wireapp/wire-server/pull/757/files#diff-851a86457730de2b9f27ae4210250ebcR1) is the nginx version that consumes this configuration. Since the work on backoffice/spar is kinda on hold, would you mind double checking whether the ciphers and TLSv1.3 are supported? (I quick research resulted in _yes_, but you probably have more knowledge around this)

The openssl version installed (1.1.1g-r0) can do TLS1.3. That's at least the version I get when installing now. Not sure what's installed on the docker image ...
I can also move that 1.3 change to a separate PR if it needs more investigation.

tiago-loureiro:
> The openssl version installed (1.1.1g-r0) can do TLS1.3. That's at least the version I get when installing now. Not sure what's installed on the docker image ...
> I can also move that 1.3 change to a separate PR if it needs more investigation.

It runs `OpenSSL 1.1.1d  10 Sep 2019` so we should be good

arianvp:
I don think backoffice serves its own TLS? isn't it terminated behind `nginx-ingress-controller` like all other services?

lucendio:
> I don think backoffice serves its own TLS so im not sure if this change does something? isn't it terminated behind `nginx-ingress-controller` like all other services?

🤔 You make an excellent point. But I guess one could argue that

a) if the image is seen as an atomic and agnostic artifact, it can run anywhere not just on an ingress terminating k8s 
b) the change affects the container-global nginx configuration. Of cource, it can be overridden by mounting another 'nginx.conf' into it's very location

I guess @franziskuskiefer was just lacking of some context and instead `grep`ing around. Which is an absolute valid approach to start.

tiago-loureiro:
> I don think backoffice serves its own TLS so im not sure if this change does something? isn't it terminated behind `nginx-ingress-controller` like all other services?

The backoffice is not "fronted" by anything, you can see more details on the [README](https://github.com/wireapp/wire-server-deploy/tree/develop/charts/backoffice)

> a) if the image is seen as an atomic and agnostic artifact, it can run anywhere not just on an ingress terminating k8s

Right, it's not running like that. 

That being said there's no good reason to have a config where undesirable ciphers/tls versions are used.

arianvp:
yeh but the backoffice currently doesn't use TLS at all; so i dont see why we even have things for it in the config. that was more my point :)

PR #329, opened 2020-08-17
Add logrotation of 3 days to all pod logs

jschaul:
Relates to https://github.com/zinfra/backend-issues/issues/1462


jschaul:
@julialongtin actually the previous settings didn't quite work, I changed them. Can you have another look? More details on the related issue.

PR #330, opened 2020-08-20
fix sanitized_request parsing on nginx

jschaul:
Related to zinfra/backend-issues#1684

tool to check regexes: https://regex101.com/


lucendio (approved):
Since https://github.com/zinfra/hegemony/pull/1671 is already merged to test on staging, I hereby approve this one. Please only merge after staging test was successful.

PR #334, opened 2020-09-03
[WIP] Use count instead of for_each because it works better

julialongtin:


lucendio:
Can you elaborate on what exactly "works better"?

In my experience, the problem with `count` is always that the *index* is being used as instance identifier and if you want to remove a single resource instance, all subsequent instances might be recreated, which might not be desired - especially with machines.

But in case this is just not applicable here or this pitfall is known already, please feel free to ignore me. I'm probably just lacking of some context.

julialongtin:
sure, that's an issue, but in this case, the use of for_each is blocking terraform from knowing how many of these instances will be created, and causing terraform not to be able to plan. so, we can either have "apply some things, with files moved around, move files back, and apply other things" with manipulated instance identifiers, OR we can have automatic identifiers, and none of the multi-staged application. this gives us the latter.

akshaymankar (approved):

PR #331, opened 2020-08-24
[tf-module:cargohold] Replace subnet IDs input with route table IDs

lucendio:
This removes the 'implicit' data dependency that can cause a race condition
between subnets and route tables. It makes it more clear what the module depends
on and should make it more easy to `depends_on` in the future (TF v0.13).

PR #323, opened 2020-08-03
Drop CBC cipher suites

franziskuskiefer:
Fixing https://github.com/zinfra/backend-issues/issues/1628

PR #336, opened 2020-09-04
Release_2020_09_04

tiago-loureiro:
# 2020-09-04

## Features

* Added support for network load balancers (#299)
* Output account number when creating gundeck resources (#326)
* Add logrotation of 3 days to all pod logs (#329)
* SFT discovery (#322)

## Bug Fixes

* Add default AWS region to example values files (#328)
* Fix sanitized_request parsing on nginx (#330)
* Updated prod-values.example with missing vars for galley (#335)

## Internal Changes

* Drop CBC ciphersuites for ingress (#323)
* Allow TLS >= 1.2 only for backoffice (#324)
* Replace subnet IDs input with route table IDs (#331)
* Explicit VPC ID when creating network load balancers (#333)


jschaul:
Obsolete. There have been fresh releases since.

PR #338, opened 2020-09-10
Remove domain name

julialongtin:
remove issuance of a default search domain via the AWS dhcp servers. breaks dns lookup inside of k8s.

### DIFF
--- a/terraform/modules/aws-vpc/main.tf
+++ b/terraform/modules/aws-vpc/main.tf
@@ -21,10 +21,6 @@ module "vpc" {
   enable_dns_hostnames = false
   enable_dns_support   = true
 
-  enable_dhcp_options      = true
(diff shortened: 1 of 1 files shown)


akshaymankar:
Why is zebot committing to this branch?

lucendio:
> Why is zebot committing to this branch?

My bet, [rebases](https://github.com/wireapp/wire-server-deploy/pull/338/commits) GH is not able to recognize as such. I have seen this before.

lucendio (approved):

akshaymankar (changes requested):
I see that those two commits are on master, but not on develop. I think @julialongtin branched off from master. IMO these two commits shouldn't go to develop.

julialongtin:
republished.

akshaymankar (approved):

PR #341, opened 2020-09-18
[sft] Add srv-announcer role to the SFT deployment automation

lucendio:
Update variables in the inventory template to match the latest version of
the role and pin to the corresponding commit.
Moved from 'roles' directive in the playbook to 'import_role' in order to
import roles instead of including them.

### DIFF
--- a/ansible/provision-sft.yml
+++ b/ansible/provision-sft.yml
@@ -9,3 +9,4 @@
 - hosts: sft_servers
   roles:
     - role: sft-server
+    - role: srv-announcer
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -66,4 +66,4 @@
 
 - src: git+https://github.com/wireapp/ansible-sft.git
   name: sft
-  version: "20d4f19685777ed3f4d11fa100bdbdd4a7e2ce1f" # master (2020-09-16)
+  version: "369db98e2e1ddf156ea0c9dbd2801c9d83e10a5a" # master (2020-09-25)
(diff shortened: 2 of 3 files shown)


ansible/provision-sft.yml
```diff
@@ -7,5 +7,8 @@
     - role: sft-monitoring-certs
 
 - hosts: sft_servers
-  roles:
-    - role: sft-server
+  tasks:
+    - import_role:
+        name: 'sft-server'
+    - import_role:
+        name: 'srv-announcer'
```
akshaymankar, line 14: Is there a difference between listing roles and importing them?

lucendio: Technically speaking, no - I remembered it differently. I thought, `roles:` would load them dynamically, but [I was wrong](https://docs.ansible.com/ansible/2.7/user_guide/playbooks_reuse.html#differences-between-static-and-dynamic). It's loading them statically. Writing it this way, it makes it more obvious, though. Anyway, I'll change it back.

ansible/requirements.yml
```diff
@@ -66,4 +66,4 @@
 
 - src: git+https://github.com/wireapp/ansible-sft.git
   name: sft
-  version: "20d4f19685777ed3f4d11fa100bdbdd4a7e2ce1f" # master (2020-09-16)
+  version: "49474b01e7e5dba3eb7e67376f8598190fd4c6cb" # move-deps-list (2020-09-24)
```
akshaymankar, line 69: This needs to be updated to point to latest master.

akshaymankar (approved):
Please only merge after changing the version in requirements.yml

PR #348, opened 2020-10-05
Use terraform outputs instead of local_file for inventory

akshaymankar:
Pre-work for: https://github.com/zinfra/backend-issues/issues/1562


lucendio (approved):
*paired on reviewing*

PR #350, opened 2020-10-06
Support keeping old SFT servers while creating new ones with differen…

jschaul:
…t server_type

This change allows making changes to the sft_server type to have a transition with servers of different types (e.g. moving from cx.11 to cx.31 without destroying or stopping exisiting servers)


jschaul:
(code co-developed with @akshaymankar and approved)

PR #352, opened 2020-10-08
update minio app and chart version

jschaul:
Relates to https://github.com/zinfra/backend-issues/issues/1659#issuecomment-705712161

### DIFF
--- a/charts/fake-aws-s3/requirements.yaml
+++ b/charts/fake-aws-s3/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: minio
-  version: 2.5.15
-  repository: https://kubernetes-charts.storage.googleapis.com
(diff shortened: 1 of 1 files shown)


lucendio (approved):
thought: if you only want to match the minio version for a specific backend installation, you could just override the image tag from a Helm `values.yaml`. But if you instead want to generally *'pin'* minio to a version compatible with `wire-server`, then this makes sense.

akshaymankar (changes requested):
Before we merge this, let us make sure our CI is able to handle downloading things from bitnami repo.

arianvp:
Superseded by https://github.com/wireapp/wire-server-deploy/pull/379

We ended up using `minio`'s chart instead of `bitnami`'s to keep closer to upstream

PR #354, opened 2020-10-12
Decrypt target

jschaul:
Markefile target to sops-decrypt the hcloud token and operator ssh key
on a given environment. I chose to not make it a prerequisite for the
other tasks, as a third-party user of this code may choose not to
encrypt these files with sops.

### DIFF
--- a/terraform/environment/Makefile
+++ b/terraform/environment/Makefile
@@ -1,7 +1,10 @@
-ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
-# Please ignore this if you're not a wire employee
+SHELL := /usr/bin/env bash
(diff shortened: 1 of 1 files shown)


terraform/environment/Makefile
```diff
@@ -39,3 +42,18 @@ else
 ENV_DIR=${CAILLEACH_DIR}/environments/${ENV}
 endif
 endif
+
+.PHONY: decrypt
+decrypt: $(ENV_DIR)/$(TOK).dec $(ENV_DIR)/$(SSH).dec
+
+$(ENV_DIR)/$(TOK).dec: check-env
```
akshaymankar, line 49: Nit: Looks like we're using round brackets and curly brackets in the same file. IIRC they mean the same things. Can we just be consistent?

akshaymankar (approved):
Nice trick! Let's also do this in `ansible/Makefile`?

PR #355, opened 2020-10-14
Add make target to get logs

akshaymankar:


ansible/Makefile
```diff
@@ -52,6 +52,27 @@ kube-minio-static-files: check-env
 		--extra-vars "service_cluster_ip=$$(KUBECONFIG=${ENV_DIR}/gen/artifacts/admin.conf kubectl get service fake-aws-s3 -o json | jq -r .spec.clusterIP)" \
 		-vv
 
+LOG_UNTIL="now"
```
lucendio, line 55: question: Did you mean to make this overridable?
suggestion:
```suggestion
LOG_UNTIL ?= "now"
```

lucendio (approved):
*after chatting about it while paring*

ansible/Makefile
akshaymankar: Thanks! Makefiles are still very confusing to me :face_with_head_bandage:

PR #356, opened 2020-10-14
Create SFT servers in two groups

akshaymankar:
This can be used for deploying new versions without taking out the previous versions.
https://github.com/zinfra/backend-issues/issues/1562
Please merge only after merging https://github.com/wireapp/ansible-sft/pull/28 and updating requirements.yaml.

### DIFF
--- a/ansible/Makefile
+++ b/ansible/Makefile
@@ -30,7 +30,7 @@ download-ansible-roles-force:
 provision-sft: check-env
 	poetry run ansible-playbook ${ANSIBLE_DIR}/provision-sft.yml  \
 		-i ${ENV_DIR}/gen/terraform-inventory.yml \
-		-i ${ENV_DIR}/inventory.yml \
+		-i ${ENV_DIR}/inventory \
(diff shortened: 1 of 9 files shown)


jschaul (approved):
Didn't try it out, but looks good.

jschaul:
So the workflow is: change list of green and blue in terraform. Apply terraform and re-generate inventory. Then run ansible. ? Maybe it would be nice to document that procedure, too.

akshaymankar:
> So the workflow is: change list of green and blue in terraform. Apply terraform and re-generate inventory. Then run ansible. ?

Slightly different. Now that we don't want to be creating/deleting instances that often. Putting things in terraform would be a one time thing (it can of course be modified later). Changing stuff on the VMs to determine which group is active and what version on SFT goes to what group would be the only things necessary before running ansible. We will add some documentation on operating SFT instances.

akshaymankar:
@jschaul I wrote some docs about operating the SFT servers. Can you please read them once before I merge this.

jschaul (approved):
Thanks for the detailed procedure!

PR #358, opened 2020-10-20
Replace shell.nix with default.nix for purity

akshaymankar:
Few major things done:
* Remove poetr2nix as it doesn't work and hides dependencies
* Add paramiko to pyproject.toml as it is needed but somehow poetry2nix provided it anyways.
* Add some missing dependencies


Dockerfile.ci-cache
```diff
@@ -0,0 +1,8 @@
+FROM nixos/nix
+
+# TODO: I am sure there is a better way to build cache than to run echo
+COPY . /wire-server-deploy
+
+RUN nix-shell /wire-server-deploy/shell.nix --arg provideCIDependencies true --command 'echo cache built'
```
akshaymankar, line 6: @flokli @arianvp This looks very roundabout, is there a better way?

flokli (approved):

Dockerfile.ci-cache
akshaymankar: Resolved by moving from `shell.nix` to `default.nix` and `nix-env`

flokli (approved):

PR #363, opened 2020-10-22
Wire offline package

flokli:
This removes poetry and uses nix to build the version of ansible we want.

It's also available in the container built with the `Dockerfile`.

This also does:
 - remove the ansible code that was downloading kubectl and helm, by using nix for it too
 - replace the playbook manually cloning the kubespray external ansible role with git submodules - given this was a workaround for broken `ansible-galaxy`and we intend to replace all this git submodules anyways.
 - update the remaining ansible roles to use git submodules too

The PR updating docs is at https://github.com/wireapp/wire-docs/pull/91.


default.nix
```diff
@@ -1,17 +1,35 @@
...
-      yq
-      gnumake
+    paths = [
+      profileEnv
+      pkgs.ansible_2_7_with_libs
+      pkgs.awscli
+      pkgs.gnumake
+      pkgs.gnupg
+      pkgs.kubectl
+      pkgs.kubernetes-helm
```
jschaul, line 29: Can we somehow version-lock the version of helm here? Currently it's on 3.3.0 it seems.

default.nix
```diff
@@ -1,17 +1,35 @@
...
-      gnupg
-      yq
-      gnumake
+    paths = [
+      profileEnv
+      pkgs.ansible_2_7_with_libs
+      pkgs.awscli
+      pkgs.gnumake
+      pkgs.gnupg
+      pkgs.kubectl
```
jschaul, line 28: hm, this loads kubectl 1.18.8 - and that is unlikely to be compatible with our clusters that are on server version 1.14 (our helm charts are known to work on 1.14 and I think 1.15 - but fail on 1.16 and above currently - see https://github.com/zinfra/backend-issues/issues/1210 and https://github.com/zinfra/backend-issues/issues/1570). Can you provide kubectl 1.14 here?

jschaul (approved):
Looks good to me. Perhaps we can find a way to version-lock kubectl and helm, but otherwise, :shipit:

nix/overlay.nix
```diff
@@ -0,0 +1,25 @@
...
+      (super.fetchpatch {
+        url = "https://github.com/ansible/ansible/pull/56903/commits/12910e828bda5da1ae9baed767fd42cb38752c19.patch";
+        sha256 = "1ba7sksmn78smsza1cp5k0bsnz6pvawj5rvmykkn88hpf2320lmf";
+      })
+    ];
+
+    propagatedBuildInputs = old.propagatedBuildInputs or [] ++ [
+      super.python37Packages.boto
+      super.python37Packages.boto3
+      super.python37Packages.six
```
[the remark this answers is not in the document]
flokli: I think most of these are already in the dependency set of ansible - see the discussion on https://github.com/zinfra/hegemony/pull/1608.

default.nix
flokli: The version lock happens implicitly via the nixpkgs pin - we could also override the src to point to another version, but hmmh…

Is there a reason we want to explicitly stickon this helm version, instead of testing this once, while introducing, and every time we bump nixpkgs?

default.nix
flokli: I think kubectl should be fine connecting to old kubernetes clusters. We can override, but I want to first test this.

arianvp: It's definitely not always fine. I burnt myself pretty badly on this before. E.g.   default between versions certain  objects get promoted to different version namespaces (E.g. in 1.18 I think `ingress` is now not in `beta` anymore). This means `kubectl` can't be used to inspect objects by default and leads to weird bugs.  (shows that there are 0 ingresses or deployments, because your deployment was deployed as `v1beta1` instead of as `apps/v1`)

For the stable apis in k8s it's never a problem; but we don't just exclusively use the stable ones.

arianvp: Version skew policies are documented here: https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl
 1.18 kubectl will not guarantee to work with 1.14

flokli: Good point. So for now we should probably provide 1.14 here.

However, we don't clearly describe which version of kubernetes we expect in `wire-server-deploy` (only in `wire-docs`, and a bit hand-wavy).

Our kubespray probably only deploys 1.14 because the playbooks default to this version. This should really be more explicit.

default.nix
arianvp: I think for helm it's fine to now implicitly depend on the version provided by the nixpkgs pin

arianvp:
Don't we also want to ship the actual playbooks in this `Dockerfile` such that the image is entirely standalone and ready to make deployments? e.g. (optionally) remove the https://github.com/wireapp/wire-server-deploy/blob/develop/Dockerfile#L9 line?

flokli:
No, this won't work. I think currently users need to update some of the group vars etc, mount their ssh keys into/generate new ones in the container etc…

I think shipping the tooling in the container, and the playbooks in either the git checkout or a tarball is fine for now.

default.nix
arianvp: Done d02ffc3dbcbfe3f5792ab5c1267a7075b2364cc2

nix/overlay.nix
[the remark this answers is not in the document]
arianvp: Lets see in a run if this works

julialongtin:
so, where is this Dockerfile located?

lucendio:
__question:__ are all python dependencies now being installed with nix instead of with poetry?

flokli:
> **question:** are all python dependencies now being installed with nix instead of with poetry?

In this PR, yes. In fact, we ship an self-contained /ansible/ thing in $PATH (both if entering via direnv or using the container image described in there).

ansible/inventory/offline/terraform-inventory.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
arianvp: Yep; no errors. will add the prelude to the thing

jschaul:
107 commits? I think this PR is getting a bit too large. How about a merge and then make more, smaller PRs for new changes?

arianvp:
But How to merge this will need some thought. Me and Florian need some breathing room to get it up to shape and maybe split it up a bit. The PR has grown in scope since we started this; as we needed a place to start with this offline deploy package.

This is mostly me and Florian stress-comitting an offline package with up to date and pinned versions of everything and fixing issues as we go with the 2 offline customers that we're currently deploying.

Once we have these customers deployed; we can take some time in elegantly splitting this up; and merging it bit by bit.  We can squash many commits together I think.

arianvp:
It's marked draft as  it's not ready for review yet. Please refrain from looking at this until we've marked it as ready. 
We're aware that in its current form it's not easy to review.  We'll squash and clean up the commit history once dust has settled a bit.

(Note that the scope of this PR has increased from what it was originally but we have not had time to reflect it in the title yet.  We're using this for integrating the work from wire-server-deploy-networkless into wire-server-deploy,  moving to a way of doing air-gap installations that is documented upstream at https://kubespray.io in the process)

ansible/offline-cluster.sh
```diff
@@ -0,0 +1,44 @@
...
+
+ansible-playbook -i ./inventory/offline ./restund.yml
+
+ansible-playbook -i ./inventory/offline ./minio.yml
+
+# Write IPs of external databases; for helm chart rendering
+ansible-playbook -i ./inventory/offline helm_external.yml
+
+# Copy ~/.kube/config from the first master node
+ansible -i ./inventory/offline "kube-master[0]" -m fetch -a  "src=/root/.kube/config dest=kubeconfig flat=yes"
```
arianvp, line 34: We can simply set `kubeconfig_localhost=true` in the inventory file and kubespray will already do this for us

ci.sh
```diff
@@ -29,6 +29,7 @@ install -m755 "$(nix-build --no-out-link -A pkgs.wire-binaries)/"* assets/binari
 function list-system-containers() {
   kubeadm config images list --kubernetes-version v1.18.10
   cat ./kubespray_additional_containers.txt
+  echo "rancher/local-path-provisioner:v0.0.18"
```
arianvp, line 32: This shouldn't be needed. It's parsed from the list of charts; no?

PR #368, opened 2020-10-26
Delete the SFT SRV record after provsioning

akshaymankar:
This will ensure that any deactivated SFT servers don't linger around due to
race between different SRV announcers.


lucendio (approved):
*paired*

PR #372, opened 2020-11-02
[ansible] Allow custom log dir when pulling logs from an instance

lucendio:
Falls back to `/tmp`


lucendio:
to little to review. *merging*

PR #373, opened 2020-11-04
Update to newer version of helm s3 plugin

jschaul:
With https://github.com/hypnoglow/helm-s3/pull/122 we can go back to the original repo and adjust our usage of the helm s3 plugin.


jschaul:
(paired with @smatting)

PR #378, opened 2020-11-13
charts/legalhold: introduce

arianvp:
A chart to install the legalhold service on-prem.

This assumes a storage provisioner is present on the cluster that
provisions volumes of storageClass: local-path

We recommend https://github.com/rancher/local-path-provisioner/tree/master/deploy/chart
which we will add in an upcoming commit

### DIFF
--- a/charts/legalhold/Chart.yaml
+++ b/charts/legalhold/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for legalhold
+name: legalhold
+version: 0.124.0
--- a/charts/legalhold/templates/deployment.yaml
+++ b/charts/legalhold/templates/deployment.yaml
@@ -1,0 +1,84 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: "{{ .Release.Name }}-hold"
+  labels:
+    name: "{{ .Release.Name }}-hold"
+  annotations:
+    checksum/secret-token: {{ include (print .Template.BasePath "/secret-token.yaml") . | sha256sum }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: "{{ .Release.Name }}-hold"
+  template:
+    metadata:
+      name: "{{ .Release.Name }}-hold"
+      labels:
+        name: "{{ .Release.Name }}-hold"
+    spec:
+      restartPolicy: Always
+      containers:
+        - name: hold
+          image: quay.io/wire/legalhold:1.0.2
+          imagePullPolicy: IfNotPresent
+
+          resources:
+            requests:
+              memory: "1500Mi"
+              cpu: 50m
+            limits:
+              memory: "1500Mi"
+
+          env:
+            - name: LD_LIBRARY_PATH
+              value: /opt/wire/lib
+
+            - name: WIRE_API_HOST
+              value: "{{ required "Must specify wireApiHost" .Values.wireApiHost }}"
+
+            - name: DELAY
+              value: 15s
+
+            - name: SERVICE_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ .Release.Name }}-hold-token"
+                  key: token
+
+            - name: DB_URL
+              value: "jdbc:postgresql://{{ .Release.Name }}-postgresql/postgres"
+
+            - name: DB_USER
+              value: postgres
+
+            - name: DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ .Release.Name }}-postgresql"
+                  key: postgresql-password
+
+          ports:
+            - containerPort: 8080
+              protocol: TCP
+            - containerPort: 8081
+              protocol: TCP
+            - containerPort: 8082
+              protocol: TCP
+
+          livenessProbe:
+            httpGet:
+              path: /status
+              port: 8080
+            initialDelaySeconds: 10
+            timeoutSeconds: 5
+            periodSeconds: 10
+
+          # for L7 LB
+          readinessProbe:
+            httpGet:
+              path: /status
+              port: 8080
+            initialDelaySeconds: 10
+            timeoutSeconds: 5
+            periodSeconds: 10
--- a/charts/legalhold/templates/service.yaml
+++ b/charts/legalhold/templates/service.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: "{{ .Release.Name }}-hold"
+spec:
+  type: ClusterIP
+  selector:
+    name: "{{ .Release.Name }}-hold"
+  ports:
+    - name: http
+      port: 80
+      targetPort: 8080
+      protocol: TCP
--- a/values/wire-server/prod-secrets.example.yaml
+++ b/values/wire-server/prod-secrets.example.yaml
@@ -63,7 +63,24 @@ nginz:
     # only necessary in test environments (env="staging"). See charts/nginz/README.md
     basicAuth: "<username>:<htpasswd-hashed-password>"
 
-team-settings:
-  secrets:
-    # Required if you want to use team-settings
-    configJson:
+# Uncomment for team-settings. Set values accordingly
+
+# team-settings:
+#   secrets:
+#     # Required if you want to use team-settings
+#     configJson: "b64 string you got from us"
+
+# Uncomment for legalhold. Set values accordingly
+
+# legalhold:
+#   serviceToken: "supersecret"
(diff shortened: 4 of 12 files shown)


charts/legalhold/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+description: A Helm chart for legalhold
+name: legalhold
+version: 0.124.0
+dependencies:
```
arianvp, line 5: Note that in helm3 we don't need a separate `requirements.yaml` anymore =)

charts/legalhold/templates/deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+  template:
+    metadata:
+      name: "{{ .Release.Name }}-hold"
+      labels:
+        name: "{{ .Release.Name }}-hold"
+    spec:
+      restartPolicy: Always
+      containers:
+        - name: hold
+          image: quay.io/wire/hold:1.0.2
```
arianvp, line 23: Yes yes I know this could be made templateable; but lets first make it work; then come up with what can all be templated.

charts/legalhold/templates/deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+            - name: DB_URL
+              value: "jdbc:postgresql://{{ .Release.Name }}-postgresql/legalhold"
+
+            - name: DB_USER
+              value: postgres
+
+            - name: DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ .Release.Name }}-postgresql"
```
arianvp, line 58: This refers to the secret that the `postgresql` chart populates.

charts/legalhold/templates/deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+              value: 15s
+
+            - name: SERVICE_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ .Release.Name }}-hold-token"
+                  key: token
+
+            - name: DB_URL
+              value: "jdbc:postgresql://{{ .Release.Name }}-postgresql/legalhold"
```
arianvp, line 50: This refers to the service that the `postgresql` chart populates

charts/legalhold/templates/secret.yaml
```diff
@@ -0,0 +1,17 @@
...
+  token: "{{ required "must specify serviceToken" .Values.serviceToken  | b64enc }}"
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "{{ .Release.Name }}-hold-tls"
+type: kubernetes.io/tls
+data:
+  tls.crt: "{{ required "Must specify tlsCrt" .Values.tlsCrt | b64enc }}"
+  tls.key: "{{ required "Must specify tlsKey" .Values.tlsKey | b64enc }}"
```
lucendio, line 17: suggestion: split these into to files (personal pref) or rename file into `secrets.yaml`
issue: add newline

charts/legalhold/templates/service.yaml
```diff
@@ -0,0 +1,14 @@
+---
```
lucendio, line 1: nitpick: consistent delimiting in LOC:1. IMO there are need for them.

values/cassandra-external/values.yaml
```diff
@@ -0,0 +1,4 @@
+IPs:
+  - 10.133.0.4
+  - 10.133.0.13
+  - 10.133.0.15
```
lucendio, line 4: issue: I think, is a leftover from bundling or test runs. Same goes for ES and minio below.

values/wire-server/prod-secrets.example.yaml
```diff
@@ -67,3 +67,95 @@ team-settings:
...
+    uGva6vpOYZIgrIYYi0okhpe8hu8mA3HwvotGCaRQoUTssrX9J62pvuRkUPKeuO4Q
+    Fx5uTMc2SHluXKtQQiqkyX5BLzYIspjd7SxDT33w2gaqptIhIl7Ip5E7XLCZ8Ro4
+    gbm7DBFcPqWoGzORMWLCDVqAFLwSBNnr9gKnCUyVrnqr0VvyWME2nVFRKZaNoX65
+    D2ypqFmOJsCLYnvHGlUoUTpzRMYEejMUvCQJK8Z9FFjI3TWh8u7qei5JhBb6BDY8
+    YPkHG4rG3DK6G4nWc4o60ig3sXoXvYct18xQhaLvGlVqvD1TZHctkHPUc3pm9MOU
+    eyIZ+E3V2fBu7ej+F6zCCHkbe6QjfyH5EHyhSWJ40fa/zj+HjV49Y73XMqVv3jIC
+    MV+1P9Fasxs1+f2fo0A1zx0ZPH2rCuZnkab0WdScmORXzZiDJfXQ9eY1M2k92s3h
+    Y6OuY6Uvi8fN54myQv5tU7Ko/nzG3yNIEW74UULB4Kh/Zb5FaBCIlRZiJosivkns
+    AjSlxx2XPfQbuybjYmkc28tp7gWA
+    -----END CERTIFICATE-----
```
lucendio, line 161: suggestion: for consistency, you may want to follow [this existing approach](https://github.com/wireapp/wire-server-deploy/blob/develop/values/nginx-ingress-services/prod-secrets.example.yaml) for placeholders.

values/wire-server/prod-values.example.yaml
```diff
@@ -1,6 +1,7 @@
 # CHANGEME-PROD: All values here should be changed/reviewed
 tags:
   proxy: false # enable if you want/need giphy/youtube/etc proxying
+  legalhold: true
```
lucendio, line 4: suggestion: default to false, since legalhold is considered evil.

lucendio (dismissed):
*Some remarks*

values/cassandra-external/values.yaml
arianvp: Thanks. I commited this by accident

values/wire-server/prod-values.example.yaml
arianvp: Will do. This was set to true so it was included in the offline-ci but I'll make a custom `values.yaml` for that

values/wire-server/prod-secrets.example.yaml
arianvp: Hmm but having placeholders that work is kind of nice.   I have double feelings about this. I agree that we should do it consistently though.  I'll remove them for now

charts/legalhold/templates/service.yaml
arianvp: I am not sure what you mean by this comment. Could you elaborate?

charts/legalhold/templates/secret.yaml
arianvp: It was originally called `secrets.yaml` but we have `secrets.yaml` in our global `.gitignore` as that's the naming convention we use for helm secret values

lucendio: Ah, right/ I remember reading about it in the chat. Splitting it would be my personal preference, but eventually, it's up to you. If you do, prefixing the file with `secret` probably makes sense.

charts/legalhold/templates/service.yaml
lucendio: I'm sorry, typos and wrong grammar.

Some of you yaml file start with `---` some don't. I suggested consistency, with the personal preference of removing it, since they are not needed. I hope, that makes the comment a bit more clear.

values/wire-server/prod-secrets.example.yaml
lucendio: > having placeholders that work is kind of nice.

Good point. But would legalhold really work with that cert and key? Another perspective could be that it might actually be desired to fail as soon as possible (maybe even during `helm install`), if both are not valid. Though, I'm not sure if that the existing approach would cause such behavior.

charts/legalhold/templates/secret.yaml
arianvp: I'll split them up in two files

charts/legalhold/templates/service.yaml
arianvp: Will do.

values/wire-server/prod-secrets.example.yaml
arianvp: So if they're not _set_ then the chart gives a useful error message at the moment (using the `required` filter in the templates) . How about I comment the sections out? So that people get a nice error message telling them they should not forget to set this value?

The certificate here is the certificate for `legalhold.example.com`.  Indeed note that useful. I'll remove the snakeoil certs and comment out the sections; so that people are clearly instructed they should set this value if they enable legalhold.

Sound good?

flokli (approved):

PR #379, opened 2020-11-13
Fix fake-aws chart on Helm 3

arianvp:
Fixes zinfra/backend-issues#1857

### DIFF
--- a/charts/fake-aws-s3/values.yaml
+++ b/charts/fake-aws-s3/values.yaml
@@ -15,5 +15,3 @@ minio:
     - name: public
       purge: false
       policy: public
-  environment:
-    MINIO_BROWSER: "off"
--- a/charts/fake-aws-s3/values.yaml
+++ b/charts/fake-aws-s3/values.yaml
@@ -5,6 +5,8 @@ minio:
   secretKey: dummysecret
   persistence:
     enabled: false
+  environment:
+    MINIO_BROWSER: "off"
   buckets:
     - name: dummy-bucket
       purge: true
--- a/charts/fake-aws-s3/requirements.yaml
+++ b/charts/fake-aws-s3/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: minio
-  version: 2.5.15
-  repository: https://kubernetes-charts.storage.googleapis.com
+  version: 8.0.3
+  repository: https://helm.min.io/


arianvp:
Am I supposed to now change the version myself? Or should CI do this after merge?

flokli (approved):

smatting (approved):

lucendio:
> Am I supposed to now change the chart version myself? Or should CI do this after merge?

[You are not](https://github.com/wireapp/wire-server-private/blob/master/ci/pipelines/wire-server-deploy-pr.yml#L176).

flokli:
So can this be merged?

arianvp:
CI is failing because of:

```
Release "test-pj1ovr3p41ui-fake-aws" does not exist. Installing it now.

Error: render error in "fake-aws/charts/fake-aws-s3/charts/minio/templates/deployment.yaml": template: fake-aws/charts/fake-aws-s3/charts/minio/templates/_helpers.tpl:49:43: executing "minio.deployment.apiVersion" at <.Capabilities.KubeVersion.Version>: can't evaluate field Version in type *version.Info

```

jschaul:
Let's get helm 3.1.1 in for CI first: https://github.com/zinfra/cailleach/pull/477

Then we can check if this is still a problem for CI, and upgrade to 3.3. One step at a time.

arianvp:
ah yeh true. seems this is mostly an egg-problem. not chicken-egg.

charts/fake-aws-s3/values.yaml
```diff
@@ -15,5 +15,3 @@ minio:
     - name: public
       purge: false
       policy: public
-  environment:
-    MINIO_BROWSER: "off"
```
jschaul, line 19: is this not needed anymore in the new chart (previously is loaded a publicly-accessible graphical user interface to poke around all the contents in minio with insecure credentials)

arianvp: It's not configurable in the new chart. At least; it complained when I had it set about the `environment` option for some reason.

It's as publicly accessible as the S3 API (it's hosted exposed on the same Kubernetes service).  I don't see more or less risk in enabling or disabling it.   Both the S3 API and the GUI use the same  AWS credentials.  If you use weak credentials then yes this is problematic. But this is orthogonal to the GUI. If exposing the Web UI is a problem then so is the API in general.

I actually found it useful for debugging to be able to access the API graphically.

jschaul (changes requested):
I just confirmed the browser is enabled by default according to https://docs.minio.io/docs/minio-server-configuration-guide - and I tried this locally, too. So please revert the removal of environment MINIO_BROWSER=off, then this PR is good to be merged, I think.

jschaul (approved):
I'd be interested if CI can deploy this. I tried locally, so it should work.

PR #380, opened 2020-11-19
Bump redis

arianvp:


arianvp:
We forgot to change https://github.com/zinfra/cailleach/blob/master/helm_vars/integration/eu-central-1/wire-server/values.yaml

I'm not sure this file should live in cailleach. We can't atomically fix CI now.  We have to break all other PRs for the duration of this PR being tested because the pipeline checks out `cailleach` master instead of a pin.  

This is a bit frustrating

flokli:
I opened https://github.com/zinfra/cailleach/pull/487, which would need to be merged, and then this pipeline be triggered.

arianvp:
Random thought. How will migration work for people bumping this chart on-prem? Does the old `redis-ephemeral` service linger around or does it get cleaned up? I'm not sure how helm handles these things

arianvp:
Rebased onto develop and ready for review

arianvp:
For the record;  `helm upgrade` did work. So my random thought can be dismissed.  (Pointed out by @lucendio in chat just now). The instructions in the release notes should be enough

lucendio (approved):
question: If I underhand this correctly, by erasing Redis from `ansible/helm_external.yml`, the scenario where Redis runs outside of a k8s is no facilitated anymore. Is that intended?

In general LGTM. Thanks for preparing the release notes!

flokli:
Yes. IIRC, was no ansible code creating external redis, and we considered it ephemeral, too.

arianvp:
https://github.com/wireapp/wire-server-deploy/pull/76 was never merged; so I think this file didn't do anything useful yet. It could be added back in that PR. We currently do not have a recommended way to do non-ephemeral redis outside of kubernetes so it looked like dead code to me.

PR #382, opened 2020-11-25
charts/sftd: introduce

arianvp:
Ran  `helm create sftd` and edited some stuff

fixes zinfra/backend-issues#1699
fixes zinfra/backend-issues#1704


charts/sftd/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
arianvp, line 18: Should be synced with all other chart versions

charts/sftd/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 1.16.0
```
arianvp, line 23: should be synced with the container image version

manifests/sftd/load-balancer/ingress.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: sftd
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    cert-manager.io/issuer: letsencrypt-staging
+    wire.com/host: sftd.k8s.arianvp.me
```
flokli, line 8: probably change that ;-)

arianvp: I'll remvove this 'experimental' code into a separate branch for now

charts/sftd/values.yaml
```diff
@@ -0,0 +1,71 @@
...
+# This must be set explicitly by the end-user. SFTD might be behind 1-1 NAT; which is common
+# in on-prem and also on GCP; so we need to explicitly tell sftd what its public IP is that
+# it can hand out to clients to open WebRTC streams
+# mediaAddress: <external IP>
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext:
+  runAsUser: 1000
```
arianvp, line 31: Note that we're not vulnerable to https://medium.com/nttlabs/dont-use-host-network-namespace-f548aeeef575 due to this `runAsUser` setting. Though it is important that the UID doesn't conflict with an existing UID on the host.  I don't know how go guarantee that.  In systemd I would just let the daemon assign one through `DynamicUser=` but kubernetes doesn't have a similar feature :(

charts/sftd/templates/statefulset.yaml
```diff
@@ -0,0 +1,89 @@
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+          args:
+            - "-I" # Address for HTTP requests
+            - "$(POD_IP)"
+            - "-M" # Address for metrics requests
+            - "$(POD_IP)"
+            - "-A" 
+            - "{{ required "Must specify mediaAddress" .Values.mediaAddress  }}"
```
arianvp, line 50: TODO: this won't work. The pod might be rescheduled and then end up on another node. This needs to be dynamic

charts/sftd/values.yaml
lucendio: Maybe we should implement the OCI spec for systemd and wrap every container runtime invocation in a service unit.

ansible/kubernetes.yml
```diff
@@ -8,4 +8,11 @@
   roles:
     - etcd-helpers
 
+- hosts: k8s-cluster
+  tasks:
+    - name: Annotate nodes
+      command: "kubectl annotate node --overwrite {{ inventory_hostname }} {{ item.key }}={{ item.value }}"
```
lucendio, line 14: thought: I'm curious if the [community.kubernetes.k8s](https://docs.ansible.com/ansible/latest/collections/community/kubernetes/k8s_module.html) collection could be used, but as long as we are on Ansible 2.7, it's not even an option.

charts/sftd/.helmignore
```diff
@@ -0,0 +1,23 @@
...
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
```
lucendio, line 23: question: is this sth we could move in to the root level of this repo?

charts/sftd/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+$ kubectl annotate node $HOSTNAME wire.com/external-ip=$EXTERNAL_IP
+```
+
+
+
+
+
+
+
```
lucendio, line 93:
~~~suggestion
$ kubectl annotate node $HOSTNAME wire.com/external-ip=$EXTERNAL_IP
```

~~~

That looks funky. Not sure if GH actually does what it's supposed to do (removing L88-L93)

File: charts/sftd/templates/ingress.yaml [hunk 9]
```diff
@@ -0,0 +1,21 @@
...
+      - "{{ required  "Must specify host" .Values.host }}"
+    secretName: "{{ include "sftd.fullname" . }}"
+  rules:
+    - host: "{{ .Values.host }}"
+      http:
+        paths:
+          - path: /
+            backend:
+              serviceName: "{{ include "sftd.fullname" . }}"
+              servicePort: sft
```
lucendio, line 21:
```suggestion
              serviceName: "{{ include "sftd.fullname" . }}"
              servicePort: sft

```

charts/sftd/templates/ingress.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: extensions/v1beta1
```
lucendio, line 1: suggestion: not sure which k8s is now the minimal required/compatible one, but if we go with a more recent one the should probably changed to `networking.k8s.io/v1`

charts/sftd/templates/service.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - port: 8585
+      targetPort: sft
+      name: sft
+    - port: 49090
+      targetPort: metrics
+      name: metrics
+  selector:
+    {{- include "sftd.selectorLabels" . | nindent 4 }}
+---
+# NOTE: I was not sure if Ingress could point to a headless service. That's why I added this
```
lucendio, line 20: question: what does headless mean?

thought: that sounds like sth that can be tested and that would eventually end up in a definitive decision on whether the second service is necessary or not.

charts/sftd/templates/statefulset.yaml
```diff
@@ -0,0 +1,119 @@
...
+            - -c
+            - |
+              set -e
+              EXTERNAL_IP=$(cat /external-ip/ip)
+              if [ -z "$ACCESS_IP" ]; then
+                ACCESS_ARGS=
+              else
+                ACCESS_ARGS="-A $EXTERNAL_IP"
+              fi
+              exec sftd  -I "${POD_IP}" -M "${POD_IP}" ${ACCESS_ARGS} -u "https://{{ required "must specify host" .Values.host }}"
```
lucendio, line 82: unsolicited comment: these are the bits where k8s really shines

charts/sftd/README.md
```diff
@@ -0,0 +1,93 @@
...
+## DNS
+
+An SRV record needs to be present for the domain name defined in `host`.
+This is an artifact of the fact the wire backend currently uses public SRV
+records for service discovery. The record needs to be of the form:
+
+```
+_sft._tcp.{{ .Values.host }}  {{ .Values.host }} 443
+```
```
lucendio, line 30: question: how does brig learn about SFT FQDNs in a k8s scenario, since there is no srv-announcer? (AFAIK hard-coded as part of the configmap, at least that is how it is with restund) 

thought: In my understanding, this part is tight to how we run SFT atm, which is w/o K8s. SO, I'd say this information is obsolete in a k8s scenario, or am I missing sth here?

lucendio (changes requested):
K8s and sftd like to allocate ports, so they might start to fight over them. Would it be necessary to put some boundaries in place. Is it even possible? Or just an unfounded concern?

charts/sftd/README.md
arianvp: Good question:

So ideally brig would just have an `sftStatic` field just like we have for TURN; that points to the loadbalancer. That code doesn't exist yet but there is a ticket in JIRA to write it.

For now; we ask operators to set up a single SRV record manually that simply points to the `A` record of the same name. That `A`  record will point to the ingress (which is loadbalanced). So there is no need to dynamically update this SRV record as it will always have just one entry.

charts/sftd/templates/ingress.yaml
arianvp: that one was introduced only in `1.19` I think. Let me double check

charts/sftd/templates/service.yaml
arianvp: A headless service is a service with `ClusterIP: None`.  Every `StatefulSet` needs an associated headless service as Kubernetes uses it to assign each pod a stable network id (https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/). 

For 1 replica the usefulness of that is less clear and it's just a ceremony we need to do for `StatefulSet` to work. But in the follow-up PR adding high availability the headless service allows us to point  the load-balancer to individual pods (e.g. to join an existing call).

https://github.com/wireapp/wire-server-deploy/pull/383/commits/a0830a8ea52bc36363d7b248b5fe144b4e0553f3#diff-15db362c68a8e0f375c147772351b59e6512d29eacfca2fde4a66e9dbeb9650fR20

Headless services don't provide a ClusterIP that does load-balancing; but gives SRV records instead. I wasn't sure if this would work with `nginx-ingress`. I'll just give it a try and see.

arianvp: Apparently ingress-nginx uses the Endpoints object directly instead of going through the `ClusterIP` to avoid an additional network hop through `kube-proxy`. So seems we can indeed remove the extra service: https://github.com/kubernetes/ingress-nginx/issues/257

Will do this now

charts/sftd/templates/ingress.yaml
arianvp: It seems to be only introduced in 1.19.     Kubespray currently uses 1.18 as its stable kubernetes version. So we can not use this yet; alas.

https://kubernetes.io/blog/2020/08/26/kubernetes-release-1.19-accentuate-the-paw-sitive/

charts/sftd/.helmignore
arianvp: `helm chart` command auto-generated this. I kept it but perhaps we can just remove this.

ansible/kubernetes.yml
arianvp: So; the offline-package branch bumps to 2.9 so we could use it there. But for now I'd just keep this as it's simple enough.

arianvp:
> K8s and sftd like to allocate ports, so they might start to fight over them. Would it be necessary to put some boundaries in place. Is it even possible? Or just an unfounded concern?

Valid concerns. but two observations:

1. sftd asks the OS for free ports; so sftd will never try to allocate a port that kubernetes is already using.
2. The only NodePort service we currently use is `ingress-nginx` which we use predefined ports for and is running as a `DaemonSet` . If `ingress-nginx` is deployed _before_ `sftd`; then everything is fine.  as `sftd` will not allocate the same port

So for people directly using our charts for configuring ingress, I think there are no issues. However this might be surprising for people deploying `sftd` in existing clusters with their own ingress controller and load-balancing setup.

A workaround could be to set  `--nodeport-addresses=127.0.0.1` on `kube-proxy` on the nodes with the `sftd` role and instruct people not to use those nodes for ingress traffic. 

But ideally; sftd should be configured to only allocate in a specific port range. This is something that has been requested and will be implemented in the future. In that case this concern will go away completely.


My suggestion is to add these notes  that I wrote here to the README  and file tickets to address these concerns in future pull requests.

charts/sftd/templates/service.yaml
arianvp: Done: https://github.com/wireapp/wire-server-deploy/pull/382/commits/37db5b8f9c6fa02bdc7f0d63e1fb98ea38c553ca

arianvp:
Update @lucendio 

After talking with @flokli he pointed out to me that the Kubernetes very much on purpose only allocates on the range `30000-32768`.

Because SFT askes the OS for ports, it will use the range defined in https://ma.ttias.be/linux-increase-ip_local_port_range-tcp-port-range/  which is `32768-62000` by default.

In other words. `sft` will never give conflictings ports with kubernetes.

charts/sftd/README.md
```diff
@@ -0,0 +1,100 @@
...
+be adjusted with the `--service-nodeport-range` flag.
+https://kubernetes.io/docs/concepts/services-networking/service/ SFTD asks the
+kernel for free ports, which by default are in the `32768-61000` range
+(https://ma.ttias.be/linux-increase-ip_local_port_range-tcp-port-range/).
+
+On a default installation these ranges do not overlap and sftd should never have
+conflicts with kubernetes components. You should however check that on your OS
+these ranges aren't configured differently.
+
+
```
lucendio, line 100:
```suggestion
these ranges aren't configured differently.

```

lucendio (approved):
tiny change request otherwise LGTM

arianvp:
This needs to wait for https://github.com/wireapp/wire-server/commit/96f4975e3f1a5b054afc4cb9ca96696fa24a0938 's CI to finish and push the new container image of brig + update the chart. And then we can merge this !

arianvp:
Scrap that. We should wait for a wire-server release; so that the chart is updated on master

arianvp:
This has been verified to work with a public IP at least!  This is ready to merge except for that https://github.com/wireapp/wire-server-deploy/pull/382/commits/544dff14a969af79758dd7b5fdd21089dcb9e8d6  should probably be dropped and we need to make a proper `wire-server` release instead...

arianvp:
Blocked by https://github.com/wireapp/wire-server-deploy/pull/393

arianvp:
I don't understand why tests are failing on this PR; as i am not touching anything `wire-server` related

PR #385, opened 2020-12-02
chrats/nginx-ingress-services: Fix invalid ObjectMeta

arianvp:
docs is not a valid components in ObjectMeta.  These objects wouldn't
apply without these changes

Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Certificate.metadat
a): unknown field "docs" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta

Ran into this whilst trying to apply this chart on my own kubernetes cluster which had cert-manager preinstalled


lucendio (approved):
Probably a result of using recent k8s a/o Helm versions.

PR #390, opened 2020-12-08
add local-path-provisioner

arianvp:
vendored from
https://github.com/rancher/local-path-provisioner/tree/master/deploy/chart,
at tag v0.0.18, 31496fdb80290f4b1425ffc3dc39b070d68a712c

Unfortunately this is not published in a helm-repository. So we use a git-submodule.

Split off https://github.com/wireapp/wire-server-deploy/pull/363

This allows us to have  `PersistentVolumes` on-prem.  local-path volumes restrict pods to be scheduled only to the node where the volume is allocated. This means less flexibility. Also it does not provide the same durability guarantees as for example Ceph.

This should be used only for components that have their own durability guarantees built-in.  In our case those are:
* minio
* elasticsearch
* cassandra

All are expected to be deployed on local disks instead of network-disks.  We don't actually have any components in our infra that don't take care of redundancy and durability themselves so local storage is very much good enough for us.

It just requires  people to be aware of the fact that it makes pod scheduling less flexible; and might need some manual intervention to get pods rescheduled when a node goes bad and needs to be replaced.  But this is no different than our current operational procedures with e.g. cassandra in our cloud.

This is a first step we can take in moving more code out of ansible and into kubernetes. Making wire easier to deploy

### DIFF
--- a/charts/local-path-provisioner/.helmignore
+++ b/charts/local-path-provisioner/.helmignore
@@ -1,0 +1,22 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/charts/local-path-provisioner/Chart.yaml
+++ b/charts/local-path-provisioner/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+description: Use HostPath for persistent local storage with Kubernetes
+name: local-path-provisioner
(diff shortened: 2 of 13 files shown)


lucendio:
question: would the plan be to publish the chart onto our chart repository?

flokli:
This is somewhat decoupled from `wire-server` as a product.

The `local-path-provisioner` chart only installs a storage provider in the cluster, but things inside wire requiring persistency, like the legalhold postgresql don't explicitly require *this* storage provider.

In a GKE cluster, there might be a provider pre-installed creating Google Compute Persistent Disks, in a vSphere cluster there might be something similar…

This is mostly to be able to provide /some/ storage provider, and vendoring it in the repo is probably sufficient. This allows to ship it in the offline package, and most likely, on an existing cluster, a more suitable provider is already installed.

We might also not only go with this one for bare-metal installations in the future, but something being a bit more reliable (in terms of "schedulable to other nodes").

julialongtin (approved):

jschaul:
Please port this PR to wire-server, as charts are no longer maintained in this repository.

PR #394, opened 2020-12-17
charts: Move to helm stable repo snapshot

arianvp:
The stable repo has been deprecated and snapshotted
https://helm.sh/docs/faq/#i-am-getting-a-warning-about-unable-to-get-an-update-from-the-stable-chart-repository

We should move away from these charts in the near future; but moving to
the snapshot suffices for now


arianvp:
I purposefully did not regenerate the lockfiles; but edited by hand; because I don't want this repo to serve us different versions

akshaymankar (approved):
:wheel_of_dharma:

PR #401, opened 2020-12-29
Remove helm charts. They moved to wireapp/wire-server

jschaul:
Helm charts were added to that repository in https://github.com/wireapp/wire-server/pull/1293

Scope: https://wearezeta.atlassian.net/browse/SQPIT-124


jschaul:
(you can ignore the CI failure: CI wants to use the helm charts, but since they are gone now, that can't work anymore. The CI pipeline has been removed now)

jmatsushita:
Hey @jschaul does this require updating docs? Seems like README in both repos and even the docs.wire.com would be affected, and probably a bunch of internal docs too?

jschaul:
> Hey @jschaul does this require updating docs? Seems like README in both repos and even the docs.wire.com would be affected, and probably a bunch of internal docs too?

This particular PR doesn't need much updating docs. In general, the related issue, https://wearezeta.atlassian.net/browse/SQPIT-124 does implicitly require updating docs a little bit. Some of that has already happened, but I agree it might be good to go over various documentation places and see what could be updated. 

docs.wire.com documentation will probably not need to change - CI keeps publishing helm charts to an S3 mirror, and that S3 mirror is what the docs on docs.wire.com make use of, so whether they are on wire-server or wire-server-deploy doesn't make a difference. Local documentation wire-server is partially there, and there's an open question here: https://github.com/wireapp/wire-server/pull/1307 on what people think would make sense to still document.

Does that answer your question?

arianvp (approved):

PR #404, opened 2021-01-25
Use nix to provide hegemony binary dependencies and switch to git submodules for ansible dependencies

arianvp:
Ths is part 1 of breaking up https://github.com/wireapp/wire-server-deploy/pull/363 into manageable chunks.

This doesn't touch any of the versions of software we use.  In follow-up PRs there will be some breaking changes


ansible/hosts.example.ini
```diff
@@ -151,11 +151,4 @@ bootstrap_os = ubuntu
 # 'flannel' is preferred on bare-metal setups, in case you wish to use `metallb`
 kube_network_plugin = flannel
 
-## You can update server-side versions for helm/kubernetes here on new releases.
-## These server-side versions should match the client-side versions:
-## See download_kubespray.yml for default server-side versions
-## See download_cli_binaries.yml for default client-side versions
-# kube_version = "v1.14.2"
-# helm_version = "v2.13.1"
-## download the kubeconfig after installing to localhost
```
lucendio, line 160: suggestion: that line can probably stay

lucendio (approved):
question: How is Helm versioned here? (that question might originate from my lack of nix knowledge). Now that I come to think of it, Helm is probably a new dependency and first introduced with this PR. Also, given the the new copy&pasta approach didn't go that far yet (deploying with Helm).

arianvp:
> question: How is Helm versioned here? (that question might originate from my lack of nix knowledge). Now that I come to think of it, Helm is probably a new dependency and first introduced with this PR. Also, given the the new copy&pasta approach didn't go that far yet (deploying with Helm).

We pin the version of nixpkgs. And nixpkgs ships with Helm 3.  I just use the latest version of helm 3 that was available there (I don't see a point to pin it to a non-upstream version in a greenfield scenario; we should always just follow the version that nixpkgs gives us when we bump nixpkgs).

ansible/hosts.example.ini
arianvp: Thanks. That last line was removed by accident

PR #405, opened 2021-01-25
Offline ansible

arianvp:


offline/docs.md
```diff
@@ -0,0 +1,339 @@
...
+d ./bin/offline-cluster.sh
+```
+
+However we now explain each step of the script step by step too. For better understanding.
+
+Copy over binaries and debs, serves assets from the asset host, and configure
+other hosts to fetch debs from it:
+
+```
+d ansible-playbook -i ./ansible/offline/inventory ansible/setup-offline-sources.yml
```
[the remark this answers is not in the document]
arianvp: Thank you!

ansible/inventory/offline/99-static
```diff
@@ -0,0 +1,145 @@
...
+#
+# Usually, you want to have a separate keypair to ssh to these boxes,
+# and tell ansible where it is by setting `ansible_ssh_private_key_file`.
+# ansible_ssh_private_key_file = ./dot_ssh/id_ed25519
+# Note adding it to the ssh agent won't work in scenarios where ansible
+# execution is wrapped through a container, as the ssh-agent socket isn't
+# accessible there.
+
+# Uncomment the following lines when you have both a public and private network
+# and you want your databases to use the private one.
```
jmatsushita, line 59: I think this needs a bit more explanation:
 - Your databases might be expanded to "your cassandra, elastic search and minio instances".
 - This is the recommended network topology no? Also the public/private terminology in an air gapped install might not be the most easy to understand? Not sure what to use instead though...
 - Also `enp1s0` might need to be changed to the name of the private interface, so we want to specify "uncomment and replace the interface name with the name of the private interface on your machines".

jmatsushita (changes requested):
Here's mostly documentation remarks, I'll start pushing some changes to address this.

offline/docs.md
```diff
@@ -0,0 +1,339 @@
+# How to install wire
+
+We have a pipeline in  `wire-server-deploy` producing container images, static
+binaries, ansible playbooks, debian package sources and everything required to
+install Wire.
+
```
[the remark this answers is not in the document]
lucendio: Would it make sense to rather have this in wire-docs? At the same time separating code and docs feels wrong to me. I'm uncertain how to go with this. Eventually, we will need to have sth over there. Although, putting (at least) some breadcrumbs here would be great.

.gitignore
```diff
@@ -17,11 +17,15 @@ values-init-done
 \#*\#
 
 # Envrc local overrides
-/.envrc.local
+.envrc.local
 
 # Nix-created result symlinks
 result
 result-*
+<<<<<<< HEAD
```
lucendio, line 25: probably not supposed to be here.
```suggestion
<<<<<<< HEAD
```

ansible/kubernetes.yml
```diff
@@ -7,11 +7,20 @@
...
+      with_dict: "{{ node_annotations | default({}) }}"
+
+- import_playbook: kubernetes_logging.yml
+
 - name: 'Bringing kubeconfig in place'
   hosts: k8s-cluster
   become: no
   tasks:
     - delegate_to: localhost
+      when: not offline
```
lucendio, line 23: Is this line still needed after #430?

offline/ci.sh
```diff
@@ -0,0 +1,121 @@
...
+docker.io/kubernetesui/metrics-scraper:v1.0.6
+EOF
+}
+
+list-system-containers | create-container-dump containers-system
+tar cf containers-system.tar containers-system
+[[ "$INCREMENTAL" -eq 0 ]] && rm -r containers-system
+
+# Used for ansible-restund role
+echo "quay.io/wire/restund:0.4.14w7b1.0.47" | create-container-dump containers-other
```
lucendio, line 62:
```suggestion
echo "quay.io/wire/restund:v0.4.16b1.0.53" | create-container-dump containers-other
```
[ ] TODO: update submodule pin of ansible-restund

arianvp: Thanks. I'll address this after rebase

offline/docs.md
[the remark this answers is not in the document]
arianvp: I think we should have it on docs.wire.com for sure.  I have had the urge to move wire-docs into wire-server-deploy before; but maybe now is not the right time

lucendio (approved):
Awesome job @arianvp

PR #406, opened 2021-01-29
Bump kubespray to 2.14.2

arianvp:


arianvp:
cc @jschaul . This will give you the same kubernetes version as the offline branch (Which I'm still slowly breaking up piece by piece).

`direnv` should have all the dependecies needed to run kubespray with kubernetes 1.18

jschaul (approved):

PR #407, opened 2021-02-02
Add MacOS to nix build matrix

arianvp:
This is just here to see that it currently doesn't work on mac


arianvp:
Build is failing because of https://github.com/NixOS/nixpkgs/pull/100717

`niv update` should fix this hopefully

PR #410, opened 2021-02-02
Upgrade required terraform version from 0.13.1 to 0.13.6

jmatsushita:
The nixpkgs update https://github.com/wireapp/wire-server-deploy/pull/408 installed terraform `0.13.6` but we requires `0.13.1`
Added github action check with `terraform init --backend=false`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,9 +23,5 @@ jobs:
           signingKey: "${{ secrets.CACHIX_SIGNING_KEY }}"
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
       - run: nix-build -A env
-  test:
-    name: test
-    runs-on: ubuntu-latest
-    steps:
-      - run: cd terraform/environment && terraform init --backend-false
-    
+      - name: Check terraform init 
+        run: cd terraform/environment && terraform init --backend-false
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -23,3 +23,6 @@ jobs:
           signingKey: "${{ secrets.CACHIX_SIGNING_KEY }}"
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
       - run: nix-build -A env
+      - run: which terraform
+      - name: Check terraform init 
+        run: nix-shell default.nix -A env --command "cd terraform/environment && terraform init --backend=false"
```
arianvp, line 28: Our `default.nix` does not work with `nix-shell` (@flokli doesn't like nix-shell. it's meant to interactively debug derivations; not for setting up `PATH`). 

Instead what you can do is:

```
export PATH=$(nix-build -A env)/bin:$PATH
( cd terraform/environment; terraform init --backend=false)
```

This is copying what our `.envrc` does in shell; basically
https://github.com/wireapp/wire-server-deploy/blob/develop/.envrc#L3

jmatsushita:
Ok @arianvp, seems good to go now.

arianvp (approved):

PR #413, opened 2021-02-03
[tf:modules,env] Factor DNS module out of hetzner-kubernetes 

jmatsushita:
NOTE: when the previous version of hetzner-kubernetes has been used before, part of the
TF state must be moved/migrated with `terraform state mv`

Additionally, environements already using this module, must set
```
sub_domains = [
  "nginz-https",
  "nginz-ssl",
  "webapp",
  "assets",
  "account",
  "teams"
]
```

dns module:
* removed var 'inject_addition_subtree' and instead indicating adding a sub-tree
  by whether 'domain' is defined or not
* default to an empty 'subdomains' instead of a pre-defined & opinionated list

instantiate DNS module:
* subdomains must explicitly be defined in environment configuration

*Co-Authored: @lucendio*

### DIFF
--- a/terraform/modules/aws-dns-records/README.md
+++ b/terraform/modules/aws-dns-records/README.md
@@ -17,14 +17,21 @@ AWS resources: route53
 module "dns_records" {
   source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-dns-records?ref=CHANGE-ME"
 
-  environment = "staging"
-
   zone_fqdn = "example.com"
+  domain = "staging"
+  sub_domains = [
+    "nginz-https",
+    "nginz-ssl",
+    "webapp",
+    "assets",
+    "account",
+    "teams"
+  ]
   ips = [ "9.9.9.10", "23.42.23.42" ]
 }
 ```
 
-If not further specified, it creates entries for the following FQDNs:
+This creates entries for the following FQDNs:
 
 * `nginz-https.staging.example.com`
 * `nginz-ssl.staging.example.com`
--- a/terraform/environment/kubernetes.dns.tf
+++ b/terraform/environment/kubernetes.dns.tf
@@ -1,0 +1,11 @@
+module "kubernetes-dns-records" {
+  for_each = toset(var.root_domain != null && length(var.sub_domains) > 0 ? [var.environment] : [])
+
+  source = "../modules/aws-dns-records"
+
+  zone_fqdn = var.root_domain
+  domain = var.environment
+  subdomains = var.sub_domains
+  ips = [ for node in local.kubernetes_nodes: node.ipaddress ]
+  create_spf_record = var.create_spf_record
+}
--- a/terraform/environment/kubernetes.dns.vars.tf
+++ b/terraform/environment/kubernetes.dns.vars.tf
@@ -1,0 +1,14 @@
+variable "root_domain" {
+  type = string
+  default = null
+}
+
+variable "sub_domains" {
+  type = list(string)
+  default = []
+}
+
+variable "create_spf_record" {
+  type = bool
+  default = false
+}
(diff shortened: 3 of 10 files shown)


terraform/environment/dns.tf
```diff
@@ -0,0 +1,29 @@
...
+    "webapp",
+    "assets",
+    "account",
+    "teams"
+  ]
+}
+
+variable "create_spf_record" {
+  type = bool
+  default = true
```
lucendio, line 15: suggestion: defaulting to `false` makes more sense, since it's mostly to facilitate the success of mail sending from the backend (`charts/demo-smtp`)

terraform/environment/dns.tf
```diff
@@ -0,0 +1,29 @@
+variable "subdomains" {
+  type = list(string)
+  default = [
+    "nginz-https",
+    "nginz-ssl",
+    "webapp",
+    "assets",
+    "account",
+    "teams"
+  ]
```
lucendio, line 10: suggestion: defaulting to this list is still implicit from the env config's perspective. I think en empty list makes sense here.

terraform/environment/dns.tf
```diff
@@ -0,0 +1,29 @@
...
+}
+
+variable "create_spf_record" {
+  type = bool
+  default = true
+}
+
+module "aws-dns-records" {
+  count =  min(1, var.kubernetes_node_count)
```
lucendio, line 19: issue: this condition seems wrong. the question is, which variable should be defined, over in the env config, that leads to this module being instantiated. Maybe `length(root_domain) > 0`? Or a simple bool flag that indicates DNS creation?

terraform/environment/dns.tf
```diff
@@ -0,0 +1,29 @@
...
+
+  source = "../modules/aws-dns-records"
+
+  environment = var.environment
+  subdomains = var.subdomains
+  # root_domain = var.root_domain
+  zone_fqdn = var.root_domain
+  ips = [ for node in local.kubernetes_nodes: node.ipaddress ]
```
lucendio, line 27: issue: I couldnt find this var anywhere? Am I holding it wrong?

thought: this seems to tighten the relationship between the hetzer-kubernetes module and this one again. Are we able to find a more generic way to obtain and feed ips into the module?

lucendio (changes requested):
General and in-line feedback:

* the READMEs of both modules need to be adjusted
* it would maybe help to give a code example in the PR description of how this would look from the the env folder's perspective

terraform/environment/dns.tf
jmatsushita: Sure, this was the default before. I'll change it.

terraform/environment/dns.tf
jmatsushita: Sure.

terraform/environment/dns.tf
jmatsushita: This was existing code in `hetzner-kubernetes`. Not sure what it's meant to do. Did you write it?

terraform/environment/dns.tf
jmatsushita: The local var is created in `inventory.tf`, as I understand it the more generic way is to use outputs but it seems we can't because of a limitation in terraform when using 2 different directory trees.

terraform/modules/aws-dns-records/README.md
```diff
@@ -17,14 +17,21 @@ AWS resources: route53
 module "dns_records" {
   source = "github.com/wireapp/wire-server-deploy.git//terraform/modules/aws-dns-records?ref=CHANGE-ME"
 
-  environment = "staging"
-
   zone_fqdn = "example.com"
+  domain = "staging"
+  sub_domains = [
+    "nginz-https",
```
arianvp, line 23: But the only place calling this module is `kubernetes-dns-records` which already defaults to the empty list; so why don't we default to the empty list here? Is it to not break existing usages?

arianvp: Discard this. I thought I was reviewing TF code but this is the README :)

arianvp (approved):
Looks good to me.

lucendio (approved):
integrated additional changes and squached everything

PR #418, opened 2021-02-08
[tf:env+module] Refactor k8s module and introduce LB

lucendio:
*Please note that this PR is best being reviewed by following the commit history*

* [x] Step 1:
  * define default machine size within the k8s module
  * remove var in env definition that is used nowhere

* [x] Step 2:
  * splitting things up into multiple files

* [x] Step 3:
  * refactor hetzner-kubernetes module:
    * Features:
      * allow to define a volume per group
      * all k8s machines are now sharing a private network
    * address FUTUREWORK
    * move from 'count' to 'for_each' in order to gain more control over which machine is created or destroyed
  * introduce new interface for the environment configuration
    * move from k8s node count to `k8s_cluster` and machine groups

#### Examples
Default, in case no k8s cluster should be defined. (Not nice, but allowing it to not exist might complicate code)
```
k8s_cluster = {
  cloud = ""
  machine_groups = []
}
``` 
A single-node instance (e.g. Wire Demo)
```
k8s_cluster = {
  cloud = "hetzner"
  machine_groups = [
    {
      group_name = "cpn"
      machine_ids = [ "1" ]
      machine_type = "cx51"
      component_classes = [ "controlplane", "node" ]
    }
  ]
}
```
Separate control plane and node groups with volumes (e.g. for etcd)
```
k8s_cluster = {
  cloud = "hetzner"
  machine_groups = [
    {
      group_name = "cps"
      machine_ids = [ "1" ]
      machine_type = "cx51"
      component_classes = [ "controlplane" ]
      volume = {
        size = 10
      }
    },
    {
      group_name = "nodes"
      machine_ids = [ "1", "2" ]
      machine_type = "cx11"
      component_classes = [ "node" ]
    }
  ]
}
```

* [x] Step 4:
  * Introduce LB
```
k8s_cluster = {
  ...
  load_balancer = true
  ...
}
```

* [x] Step 5:
  * refactor DNS module to support defining different IPs for SPF


arianvp:
Having to manually specify a list machine id's feels like a smell to me. Can't we just generate some with https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/pet  and `count` ?

I only care about the size of the node-pool; I should never care about the names of the nodes in a node-pool in kubernetes.

arianvp:
Second remark:

```
k8s_cluster = {
  cloud = "hetzner"
  machine_groups = [
    {
      group_name = "cpn"
      machine_ids = [ "1" ]
      machine_type = "cx51"
      component_classes = [ "controlplane", "node" ]
    }
  ]
}

```

I don't know if I would put this much freedom at the vars level. I'd push some of these concerns down to the terraform code.  We're now at a point we're sort of duplicating a subset of the terraform DSL in vars.  We're giving a bit too much freedom in that it allows for configurations that might not work in practice.  E.g. with etcd  it only works in uneven configurations.  `1`, `3` or `5` nodes.  But the variables give us the freedom to configure etcd with 2 nodes. This is not ideal.

l. I would just hardcode at the terraform level that we have `3` control-plane nodes and only make configurable whether we consider them as a worker pool or not.
```hcl
resource "hetzner_server" "controlplane" {
   count = 3
   name = "controlplane${count.index}"
   labels = { "component-classes.wire.com/control-plane": true,  "component-classes.wire.com/node": vars.controlplane_is_worker }
}
```

I can get behind being able to define multiple node pools in vars though. That sounds useful (due to the lack of autoscaling groups in hetzner)

lucendio:
> Having to manually specify a list machine id's feels like a smell to me. Why can't we use `count` here?

No good reason. You are right here. Especially since the `group_name` still allows to life cycle machines by life cycling the whole group. We were already talking about this aspect when we discussed the schema initially. It just happen that I forgot about this, and still had the old way (`machine_ids`) in the test configuration.

A Note on exposing this kind of level in the environment configuration: in a follow-up verbal discussion, we decided  to go for this for now, migrate the the different configurations across all environments to the current schema and then reiterate on the overall approach by looking into the idea of having some kind of thin TF layer in the form of opinionated *'template modules'* which could then leverage the modules doing the heavy lifting and in turn would look very similar (or maybe even simpler) compared to the `k8s_cluster` schema. Downside of this, an environment would start to contain again some actual TF code.

ansible/kubernetes.yml
```diff
@@ -1,6 +1,31 @@
...
+      - when: file_adminconf.stat.exists
+        block:
+        - name: 'Renaming kubeconfig file provided by Kubespray'
+          copy:
+            src: "{{ artifacts_dir }}/admin.conf"
+            dest: "{{ artifacts_dir }}/../kubeconfig.dec"
+        - file:
+            path: "{{ artifacts_dir }}/admin.conf"
+            state: absent
+
```
lucendio, line 28: @arianvp this probably breaks an assumption for in the offline package, doesnt it? 

Note, this is the first step (out of two) to automate sopsing kubeconfig in an $ENV_DIR when bootstrapping a K8s cluster.

terraform/modules/hetzner-kubernetes/machines.variables.tf
```diff
@@ -0,0 +1,35 @@
...
+
+# FUTUREWORK: replace 'any' by implementing https://www.terraform.io/docs/language/functions/defaults.html
+#
+variable "machines" {
+  description = "list of machines"
+  # type = list(object({
+  #   group_name = string
+  #   machine_id = string
+  #   machine_type = string
+  #   component_classes = list(string)
```
akshaymankar, line 22: As I understand there are 2 valid component classes: "controlplane" and "node" and terraform provides no way of doing enums. So, can we do the next best thing which would be to document and validate this?

I have a feeling that we are over-generalizing here. Nobody wants a "machine" in their kubernetes cluster. Maybe it is better to accept "pools" for nodes and controlplane separately. This will also get rid of the enum problem.

terraform/environment/kubernetes.cluster.vars.tf
```diff
@@ -0,0 +1,29 @@
...
+  #     component_classes = list(string)
+  #     volume = optional(object({
+  #       size = number
+  #       format = optional(string)
+  #     }))
+  #   }))
+  # })
+  type = any
+  default = {}
+}
```
akshaymankar, line 29: Why do we want to encode all kinds of k8s_clusters in one type of specification? I have a feeling that this will bite us in future as we might get into specifying AWS specific things somehow. Maybe it is better to just have this variable be called "hetzner_k8s_cluster". When we start providing an option to support AWS, we can just create another variable called "aws_k8s_cluster"?

terraform/environment/kubernetes.inventory.tf
```diff
@@ -0,0 +1,35 @@
...
+          {
+            ansible_host = m.public_ipv4
+            ip = m.private_ipv4
+          },
+          contains(keys(m), "etcd_member_name" ) ? { etcd_member_name = m.etcd_member_name } : {}
+        )
+      }
+      vars = merge(
+        {
+          # NOTE: please consider setting variables in the inventory of the respective environment instead
```
akshaymankar, line 26: I maybe not have enough context, but I cannot figure out what this comment means?
Also, why is merging with an empty map required?

terraform/environment/kubernetes.cluster.vars.tf
akshaymankar: We should also ensure that we don't create even number of etcds as it provide the same fault tolerance as creating 1 less etcd.

terraform/modules/hetzner-kubernetes/load-balancer.resources.tf
```diff
@@ -0,0 +1,80 @@
...
+  count = var.with_load_balancer ? 1 : 0
+
+  name     = "${var.cluster_name}-lb"
+  location = var.default_location
+
+  load_balancer_type = (
+    length(var.machines) <= 5 ? "lb11" : (
+    length(var.machines) <= 75 ? "lb21" : (
+    length(var.machines) <= 150 ? "lb31" : null))
+  )
```
akshaymankar, line 11: I see that this is a required thing which could potentially be null, does terraform plan fail in this case? What error would the user get?

akshaymankar (commented):
I dropped some questions inline.
Apart from that I have one question: What is the migration path for the old environments? Do we recreate them or there is some terraform state migration that we have to do?

terraform/modules/hetzner-kubernetes/machines.variables.tf
arianvp: This seems to be the same concern as I expressed here https://github.com/wireapp/wire-server-deploy/pull/418#issuecomment-776606347

My biggest concern is that we are giving too much freedom. E.g. nothing is stopping you from accidentally assigning `2` or `4` control-plane nodes; whilst those setups do not work in etcd.  you should only have `1` or `3` control-plane nodes and _in very rare cases_ where you have many regions or zones, an etcd cluster of `5` is possible (but has bad performance. As every write needs consensus of 4 nodes instead instead of 2  )


I would still suggest us having two sections:

```
control_plane =  {
  count  = 3  # can be 3 or 1; but no other value
  is_worker = true  # This sets the component_class label on the control_plane nodes
}

node_pools = [
{ count = 200 name = "basic",  machine_type = blah }
{ count = blah,  machine_type = blah }
]
```

terraform/modules/hetzner-kubernetes/load-balancer.resources.tf
lucendio: It fails with

```
Error: Required attribute is not set

  on ../modules/hetzner-kubernetes/load-balancer.resources.tf line 7, in resource "hcloud_load_balancer" "lb":
   7:   load_balancer_type = null
```

This is IMHO sufficiently self-explanatory after looking at the corresponding LOCs. Please yell if you disagree!

terraform/modules/hetzner-kubernetes/load-balancer.resources.tf
```diff
@@ -0,0 +1,80 @@
...
+
+  # NOTE: not sure what impact the fact has that one LB targets two possibly
+  #       disjunct groups of machines and what role the algo play in that
+  algorithm {
+    type = "round_robin"
+  }
+}
+
+
+resource "hcloud_load_balancer_network" "lb-nw" {
```
arianvp, line 21: I think this should preferably not live in the kubernetes module, but in a separate `private_network` module and then be an input to the kubernetes module.

Say I have another module defining `cassandra` cluster or `elasticsearch` cluster; I want these to be in the same private network as the kubernetes cluster.  

Another option would be to let the `kubernetes` module provide the private network as an output, but this is harder for refactoring and a bit awkward (Why would the `cassandra` module have to depend on the `kubernetes` module; for example)

See Also See the section about [Dependency Inversion](https://www.terraform.io/docs/language/modules/develop/composition.html#dependency-inversion) in the modules best practises guide.

arianvp: Scrap this comment I put it on the wrong part of the file

terraform/modules/hetzner-kubernetes/machines.resources.tf
```diff
@@ -0,0 +1,54 @@
...
+      group_name = each.value.group_name
+      machine_id = each.value.machine_id
+    },
+    contains( each.value.component_classes, "controlplane" ) ? { etcd_member_name = "etcd-${ each.value.machine_id }" } : {},
+    { for class in each.value.component_classes : "component-class.${local.LABEL_PREFIX}/${class}" => true }
+  )
+}
+
+
+resource "hcloud_server_network" "snw" {
```
arianvp, line 24: I think this should preferably not live in the kubernetes module, but in a separate `private_network` module and then be an input to the kubernetes module.

Say I have another module defining `cassandra` cluster or `elasticsearch` cluster; I want these to be in the same private network as the kubernetes cluster.  (Because I want kubernetes nodes to be able to talk to these services)

Another option would be to let the `kubernetes` module provide the private network as an output, but this is harder for refactoring and a bit awkward (Why would the `cassandra` module have to depend on the `kubernetes` module; for example)

See Also  the section about [Dependency Inversion](https://www.terraform.io/docs/language/modules/develop/composition.html#dependency-inversion) in the modules best practises guide.

terraform/modules/hetzner-kubernetes/machines.resources.tf
```diff
@@ -0,0 +1,54 @@
...
+  for_each = toset([ for m in var.machines: "${m.group_name}-${m.machine_id}" ])
+
+  server_id = hcloud_server.machines[each.key].id
+  subnet_id = hcloud_network_subnet.sn.id
+}
+
+
+resource "hcloud_volume" "volumes" {
+  for_each = {
```
arianvp, line 33: What do we need statically defined volumes for? If we need storage in kubernetes, we should do that through `PersistentVolume` abstraction and the [csi-driver](https://github.com/hetznercloud/csi-driver).

terraform/modules/hetzner-kubernetes/load-balancer.resources.tf
arianvp: Hmm it is a bit scary though. Say I bump my cluster from 5 to 6 nodes; that will get the load-balancer destroyed and recreated and cause a period of downtime. This sounds less than ideal. Wouldn't this better be something we should bubble up and let the administrator make a concious decision about?

terraform/modules/hetzner-kubernetes/machines.resources.tf
lucendio: The scenario you are describing was exactly @jmatsushita and my reasoning behind introducing `machine_groups`. This allows to create only K8s groups (controlpales & nodes) but also other groups in addition, by deriving the Ansible groups from whatever items are defined in `component_classes`. Meaning a item of `component_classes` equals to an Ansible inventory group. This way, all machines would share the same private network.

terraform/modules/hetzner-kubernetes/machines.resources.tf
lucendio: That was a [FUTUREWORK](https://github.com/wireapp/wire-server-deploy/blob/develop/terraform/modules/hetzner-kubernetes/main.tf#L1) I just tried to address.

terraform/modules/hetzner-kubernetes/load-balancer.resources.tf
lucendio: You absolutely have a point here, but before drawing any conclusion, I think we should test whether it actually causes destruction and downtime. Additionally, that the LB might get destroyed should be completely transparent in the TF plan. Wouldn't this suffice as a heads-up and pull back if necessary?

arianvp: Yes I suppose it does.

terraform/modules/hetzner-kubernetes/machines.resources.tf
arianvp: Hmm I really feel like we're trying to re-invent the wheel here a bit too much still.  What doe we gain from using `machine_groups` in `tfvars` over defining a `resource` (or a `module`)  in terraform code? We're coming up with our own DSL for defining machine groups; which terraform itself already has...    Defining how we provision `cassandra` and elasticsearch groups  whilst calling the `kubernetes` module also sounds somewhat confusing to me.    I am still not really fond of this idea.

arianvp: Why I am not fond of it:

The terraform code alone will not describe the topology of a Wire installation anymore. There will just be one node in the graph called "machine" and it's up to the user to define a topology in their `tfvars` that makes sense.

If we instead have explicit `cassandra`, `kubernetes` and `elasticsearch` `hcloud_server` resources (or modules if there is a group of resources you want to abstract) then it's clear from the terraform code what a wire deployment topology looks like. 

This means we have to explicitly document how people should configure this topology in `tfvars` instead of having it evident in the code itself.

terraform/modules/hetzner-kubernetes/machines.resources.tf
arianvp: As we're already experiencing disk latency issues on the local disks on hetzner every once in a while (or at least that's the running theory) I'm not sure it's a good idea to run the etcd data directory on a network disk .

Usually replacing an etcd node with an empty node will just work; as it will backfill from the other server nodes.   (When the data is large; it can help copying over data from the old node first before killing it; and then copying it to the new node; but we can do that manually)

terraform/modules/hetzner-kubernetes/load-balancer.resources.tf
arianvp: I'm convinced :) Consider this resolved

lucendio:
After a verbal discussion, we decided to merge as is and address the following remaining issues in a follow-up PR (https://github.com/wireapp/wire-server-deploy/pull/423), because with a slight change of design it will straight-forward to make a *kubernetes-hetzner* module more robust and specialized.

* [ ] validate etcd (or only give specific choices) according to the [Raft toleration behaviour](https://etcd.io/docs/v3.3.13/faq/#what-is-failure-tolerance)
* [ ] change interface design according to the [suggestion](https://github.com/wireapp/wire-server-deploy/pull/418#discussion_r577643963)
* [ ] decide whether to keep volume creation in there
* [ ] factor out private network creation

arianvp:
The CI failure is fixed in https://github.com/wireapp/wire-server-deploy/pull/424

arianvp (approved):

PR #419, opened 2021-02-15
Add sftd configuration to wire-server values

arianvp:
Depends on https://github.com/wireapp/wire-server/pull/1325


arianvp:
Honestly it would be a lot easier if we'd move the `values.yaml` to `wire-server`. Would require changes to https://docs.wire.com Thoughts about that?

arianvp:
Technically not needed anymore because I disabled `sftd` by default in the `wire-server` chart... For now...

PR #420, opened 2021-02-16
fix wire-server-metrics example values namespace

tjanson:
The keys being nested under `wire-server-metrics` in the example values file for that chart was a typo / probably a leftover from some other nested chart.


lucendio (approved):
Thank you!

tjanson:
My bad, the nested keys are actually still wrong, in more than one way. I'll add a PR.

PR #421, opened 2021-02-17
Fix demo values

arianvp:
Without these values the webapp would not work 

Fixes https://wearezeta.atlassian.net/browse/SQPIT-92


values/wire-server/demo-values.example.yaml
```diff
@@ -165,7 +165,39 @@ webapp:
       backendRest: nginz-https.example.com
       backendWebsocket: nginz-ssl.example.com
       backendDomain: example.com
+      backendTeamSettings: teams.example.com
```
lucendio, line 168:
```suggestion
      # When 'tags.team-settings: true'
      backendTeamSettings: teams.example.com
```

lucendio (approved):

PR #422, opened 2021-02-17
Bump restund

arianvp:
Needed for offline. Drops `rkt` support.

Have not tested the migration path; as the current playbook on `develop` does not even work


lucendio (approved):
Although it might be obvious to some (submodule ref change), would you mind putting a link to the breaking changes (i.e. https://github.com/wireapp/ansible-restund/commit/4db0bc066ded89cf0ae061e3ccac59f3738b33d9)

PR #424, opened 2021-02-22
Move to wire-server-deploy cachix cache

arianvp:
The old cache is gone (was on my private account. now using Wire's account)


lucendio (approved):
Let's see.

PR #426, opened 2021-02-24
Set `access_ip` for machines as well

arianvp:
Otherwise etcd will try to form a cluster over the public network which e.g. won't work in offline and also `etcd` benefits from the lower latencies.


arianvp:
Nope sorry I think I misconfigured something else.

PR #427, opened 2021-02-24
Add srv to dns module

jschaul:
For federation, we will in a first phase (subject to change later) use SRV records for server discovery between backends. For that, we need SRV record support in this module and the kubernetes environment.

### DIFF
--- a/terraform/modules/aws-dns-records/resources.route53.tf
+++ b/terraform/modules/aws-dns-records/resources.route53.tf
@@ -28,20 +28,20 @@ resource "aws_route53_record" "spf" {
   ttl     = "60"
   records = [
     join(" ", concat(
-      [ "v=spf1" ],
-      [ for ip in var.spf_record_ips : "ip4:${ ip }" ],
-      [ "-all" ]
+      ["v=spf1"],
+      [for ip in var.spf_record_ips : "ip4:${ip}"],
+      ["-all"]
     ))
   ]
 }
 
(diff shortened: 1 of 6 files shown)


terraform/modules/aws-dns-records/resources.route53.tf
```diff
@@ -28,9 +28,20 @@ resource "aws_route53_record" "spf" {
...
+
+resource "aws_route53_record" "srv-server" {
+  for_each = var.srvs
+
+  zone_id = data.aws_route53_zone.rz.zone_id
+  name    = join(".", concat([each.key], local.name_suffix))
+  type    = "SRV"
+  ttl     = "60"
+
+  records = [for t in each.value : join(".", concat([t], local.name_suffix))]
```
lucendio, line 46: note: that one gave me quick headache. clever, tho, but a bit confusing too (when looking at the example). Would some interpolation syntax make it more obvious?

lucendio (approved):
LGTM

terraform/modules/aws-dns-records/resources.route53.tf
jschaul: interpolation would probably be better. the `join(concat` is confusing. But currently local.name_suffix is a list. We could refactor this though to become `[for t in each.value : "${t}.${local.domain_suffix}`. I'll make a follow-up PR for that.

PR #428, opened 2021-02-25
update bin/secrets.sh to aid in creating fresh environments with fres…

jschaul:
…h secrets

Motivation: as I was creating new environments, I needed to create fresh secrets for them, and add them to a secrets.yaml file. This is currently a tedious process involving copy-pasting the right bits of string to the rightly-indented json.

If this PR is accepted, FUTUREWORK could be to A) add the existence of this script to docs.wire.com and B) we could make it even easier by providing a docker image containing zauth, openssl, and htpasswd.

### DIFF
--- a/bin/secrets.sh
+++ b/bin/secrets.sh
@@ -11,6 +11,7 @@ TOPLEVEL_DIR="$( cd "$SCRIPT_DIR/.." && pwd )"
 # ./bin/secrets.sh [ path-to-new-directory-for-secrets-output | default ./secrets_cache ]
 
 OUTPUT_DIR="${1:-"$TOPLEVEL_DIR/secrets_cache"}"
+ZAUTH_CONTAINER="${ZAUTH_CONTAINER:-quay.io/wire/zauth:latest}"
 
 mkdir -p "$OUTPUT_DIR"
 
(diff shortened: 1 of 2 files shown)


bin/secrets.sh
```diff
@@ -1,15 +1,29 @@
...
-zrest="${ANSIBLE_DIR}/secrets/restund_zrest_secret.txt"
-zpub="${ANSIBLE_DIR}/secrets/zauth_public.txt"
-zpriv="${ANSIBLE_DIR}/secrets/zauth_private.txt"
-miniopub="${ANSIBLE_DIR}/secrets/minio_public.txt"
-miniopriv="${ANSIBLE_DIR}/secrets/minio_private.txt"
+OUTPUT_DIR="${1:-"$TOPLEVEL_DIR/secrets_cache"}"
+
+mkdir -p "$OUTPUT_DIR"
+
+zrest="${OUTPUT_DIR}/restund_zrest_secret.txt"
```
arianvp, line 15: I actually wrote the _exact same script_ yesterday https://github.com/wireapp/wire-server-deploy/pull/405/commits/8f9683fc4f846ed0b7f05f9fcb55e8a5690e294a

:) 

However I removed the writing to `.txt` files I wrote the secrets directly to the ansible inventory instead (through a `group_vars` file).

Perhaps that's also something we want to do here? Then you don't have to copy the secrets into your ansible inventory manually

jschaul: Ah well we could have saved some time by not doing that concurrently. Oh well :)

I took your ansible-inventory-generating code from the linked commit and added it here. Thanks! I also took the ZAUTH_CONTAINER idea and integrated it here.

As for the question of saving to text file: well, this PR also adds a basicAuth secret, which needs to be shared via group password manager and is neither used by helm nor by ansible. So at least that one needs its own file. Storing individual secrets in a file makes it easier to re-generate something or change the script. Do you think it's necessary/better to remove the writing-one-secret-to-file?

arianvp (approved):
This looks good!  Would you want to add `openssl` and `apacheHttpd` to the `default.nix` file?  This will make `htpasswd` and `openssl`  CLI available for people with `direnv` and for people using the `wire-server-deploy` docker container (e.g. in offline deploys)

jschaul:
> This looks good! Would you want to add `openssl` and `apacheHttpd` to the `default.nix` file? This will make `htpasswd` and `openssl` CLI available for people with `direnv` and for people using the `wire-server-deploy` docker container (e.g. in offline deploys)

Will do in a follow-up PR!

PR #430, opened 2021-03-01
[ansible] Provide backward compatibility for offline pkg

lucendio:
This way the original admin.conf stays where it is, but at the same time kubeconfig
is created (or at least prepared) in the env dir root, ready to be sopsed.


arianvp (approved):
Thanks this is useful :)

PR #431, opened 2021-03-02
Add apacheHttpd and openssl to default.nix

smatting:


jschaul (approved):

lucendio:
What are the these two dependencies for if I may ask?

smatting:
Pinging @jschaul here, because we added these in a parining session

jschaul:
The bin/secrets.sh script makes use of them to generate secrets.

PR #433, opened 2021-03-04
Add helm plugins

jmatsushita:
With https://github.com/NixOS/nixpkgs/pull/106901 merged, install needed helm plugins with nix, and also throwing `rke` in here, as I work on https://wearezeta.atlassian.net/browse/SQPIT-62


arianvp (approved):

jmatsushita:
Removed `rke` client as it's not needed with `rke2`

default.nix
```diff
@@ -15,28 +15,30 @@ let
     '';
   };
 
+  helmWithPlugins = pkgs.wrapHelm pkgs.kubernetes-helm {
+    plugins = with pkgs.kubernetes-helmPlugins; [ helm-s3 helm-secrets helm-diff ];
```
flokli, line 19: This feels a bit redundant with the helm plugin stuff in cailleach. Can we agree where to put this? ;-)

PR #438, opened 2021-03-24
Dockerfile: Use cachix cache

akshaymankar:


arianvp:
Hmm I think I'd like it better if we hardcode the public key inside the nix config instead. This way cachix can not MITM us.

akshaymankar:
> Hmm I think I'd like it better if we hardcode the public key inside the nix config instead. This way cachix can not MITM us. Blindly trusting cachix to give us the correct public key kind of defeats the point of signing the builds locally in the first place.

I have not clue how to do this, haven't tried to find it out yet. But it doesn't look like I will get to it this week, so I will close this and feel free to do this in a separate PR.

PR #443, opened 2021-04-01
Fix ansible module dependencies

arianvp:
What we were doing before was way too complicated. Ansible itself
doesn't have any dependency on boto to function. The _target host_
needs to have boto installed.

In our case the target host is 'localhost'.  Ansible will by default
pick /usr/bin/python3 as the python interpreter ogit n target hosts but of
course we can't rely on that having boto installed.

Instead, we configure ansible_python_interpreter to point to the
environment we built with nix; which contains a python that has boto
installed.

This way boto can be found and the playbooks should succeed!

### DIFF
--- a/ansible/host_vars/localhost/python.yml
+++ b/ansible/host_vars/localhost/python.yml
@@ -1,1 +1,1 @@
-ansible_python_interpreter: $LOCALHOST_PYTHON
+ansible_python_interpreter: "{{ lookup('env','LOCALHOST_PYTHON') }}"
--- a/ansible/host_vars/localhost/python.yml
+++ b/ansible/host_vars/localhost/python.yml
@@ -1,0 +1,1 @@
+ansible_python_interpreter: $LOCALHOST_PYTHON
--- a/.envrc
+++ b/.envrc
@@ -1,11 +1,13 @@
-env="$(nix-build $PWD/default.nix -A env --no-out-link)"
+nix-build $PWD/default.nix -A env --out-link .nix-env
 
-PATH_add "${env}/bin"
+PATH_add ".nix-env/bin"
+
+export LOCALHOST_PYTHON="$PWD/.nix-env/bin/python"
 
 # source .profile from `$env`.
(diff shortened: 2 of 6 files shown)


lucendio (approved):
LGTM but DH action for macos does not agree.

arianvp:
This fixes it locally but it's still broken in the dockerfile. I'll try to get that to work Tuesday first

arianvp:
I tested this with the following playbook:

```ansible
- hosts: localhost
  gather_facts: yes
  serial: 1
  become: no
  tasks:
    - debug: var=ansible_python_interpreter
    - debug: var=ansible_python_version
    - debug: var=ansible_playbook_python
    - aws_s3:
        mode: list
        bucket: public.wire.com
```

ansible/host_vars/localhost/python.yml
```diff
@@ -0,0 +1 @@
+ansible_python_interpreter: $LOCALHOST_PYTHON
```
lucendio, line 1: For me, this fails with:
```
PLAY [localhost] **********************************************

TASK [Gathering Facts] **********************************************
fatal: [localhost]: FAILED! => {"ansible_facts": {}, "changed": false, "failed_modules": {"setup": {"failed": true, "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 0}}, "msg": "The following modules failed to execute: setup\n"}

PLAY RECAP **********************************************
localhost                  : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

What you probably want to do is
```suggestion
ansible_python_interpreter: "{{ lookup('env','LOCALHOST_PYTHON') }}"
```
or maybe even add a default in addition, if you think it makes sense:
```
ansible_python_interpreter: "{{ lookup('env','LOCALHOST_PYTHON') | default('/some/path') }}"
```

lucendio (changes requested):

lucendio (approved):

PR #444, opened 2021-04-09
Fix submodules defining  ansible-kubectl

lucendio:


lucendio:
@arianvp what about `ansible-helm`

arianvp:
@lucendio yes ansible-helm is also unused as far as I know

arianvp (approved):

ansible/Makefile
```diff
@@ -43,7 +43,7 @@ ifndef LOG_SINCE
 endif
 	ansible-playbook ${ANSIBLE_DIR}/get-logs.yml \
 		-i ${ENV_DIR}/gen/terraform-inventory.yml \
-		-i ${ENV_DIR}/inventory.yml \
```
arianvp, line 46: Just checking. is the dropping of `.yml` intentsional herE?

lucendio: the whole change was not meant to be in here. Will remove it.

PR #445, opened 2021-04-09
[submodules] Bump sft role to v0.4.0

lucendio:
Otherwise we might run into issues with its subsequent dependency.
See https://github.com/cloudalchemy/ansible-node-exporter#warning


arianvp:
We have this role as a submodule; no? Do we have to change the naming scheme in there as well? 

https://github.com/wireapp/wire-server-deploy/blob/develop/.gitmodules#L58-L60

lucendio:
> We have this role as a submodule; no? Do we have to change the naming scheme in there as well?
> 
> https://github.com/wireapp/wire-server-deploy/blob/develop/.gitmodules#L58-L60

nope. the issues only occurs when pulling it from Ansible Galaxy.

arianvp (approved):

PR #446, opened 2021-04-12
[submodules] Bump ansible-restund to v0.2.6

lucendio:
Question: setting some CIDRs for the Restund firewall, do we want/need to somehow include this in the offline installation procedure?


arianvp (approved):

arianvp:
Why wouldn't we want to include it? 

We should however perhaps document how to add the exceptions to the `group_vars` if people are using certain local ranges to connect to restund from clients

PR #432, opened 2021-03-03
Concourse deploy tooling

jmatsushita:
The main focus of this PR is to pass etcd information since `kubespray` installs it outside of kubernetes, so that the prometheus operator can access the etcd certificates and ip addresses. https://github.com/zinfra/cailleach/pull/521/files#diff-51b29e192d0e2c19444e320157af78ca1adeda1fcda531a30846508f11449973R37-R40

Here are a few thigns I found useful while setting up the concourse kubernetes cluster:
 - ensure the gen folder is destroyed on `make -C terraform/environment destroy EMV=ci-dev` 
 - targeted playbook for ansible with `make -C ansible play-helm_etcd ENV=ci-dev` for example.
 - targeted helmfile apply with `make -C helm apply-concourse-ci ENV=ci-dev`

PR #447, opened 2021-04-14
[tooling] Merge [Terraform,Ansible,Helm] Makefile user interfaces

lucendio:
To prevent duplication, simplify maintainability and prevent version drift, this
change set aims to merge existing Makefiles into one located at the root of the
code base.

The resulting Makefile introduces a concepts of target dependencies (see `check-*`),
which are certain files or folders put in place by a previous step (e.g. Terraform
generates an inventory consumed by Ansible targets). Thus, running `make decrypt`
is still required, but the `check-inputs-*` will check whether the decrypted version
already exist, and errors if not.

* adds asserts to log extraction playbook to get rid of native make conditions; set
  local default location to ENV_DIR
* the necessity of setting ENV or ENV_DIR did not change; it's just that the check
  has been moved to the top of the Makefile. This way make fails as early as possible
* targets abstracting Terraform invocations now require the '.terraform' folder to
  exist, which implies `terraform init` to be invoked upfront, if '.terraform/' is
  not already there. Also, `make init` was renamed to `make re-init` due to the
  described implicit behaviour.

### Follw-up Tasks
* [ ] adjust CICD code to use the new Makefile
* [ ] adjust READMEs
* [ ] pick and migrate from https://github.com/wireapp/wire-server-deploy/pull/432
* [ ] remove the original Makefiles

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,14 +35,6 @@ $(TF_DATA_DIR):
 re-init: check-inputs-terraform
 	make --always-make $(TF_DATA_DIR)
 
-# NOTE: alias for 're-init'
-.PHONY: init
-.SILENT: init
-init:
-init:
-	echo "[WARN] 'make init' is deprecated. Please use re-init moving forward"
-	make re-init
-
 .PHONY: apply plan console destroy
 apply plan console destroy: check-inputs-terraform $(TF_DATA_DIR)
 	cd $(TERRAFORM_WORKING_DIR) \
@@ -160,7 +152,7 @@ kubeconfig.dec: check-env-dir
 		test -s $(ENV_DIR)/$(@) || (echo "[ERR] Decryption failed: $(basename $(@))" && exit 1) \
 	)
 
-.PHONY: clean
+.PHONY: clean-decrypt
 clean-decrypt: check-env-dir
 	rm $(ENV_DIR)/*.dec
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,201 @@
+SHELL := /usr/bin/env bash -eo pipefail
+MKFILE_DIR := $(abspath $(dir $(abspath $(firstword $(MAKEFILE_LIST)))))
+
+# NOTE: variables that can be defined/overwritten
+#ENV_DIR or ENV
+
+# NOTE: internal variables
+# Please ignore the following line if you're not a Wire employee
+ENVIRONMENTS_DIR := $(abspath $(CURDIR)/../cailleach/environments)
+
+
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,201 @@
...
+
+.PHONY: re-init
+re-init: check-inputs-terraform
+	make --always-make $(TF_DATA_DIR)
+
+# NOTE: alias for 're-init'
+.PHONY: init
+.SILENT: init
+init:
+init:
```
jschaul, line 42: duplicate line?

Makefile
```diff
@@ -0,0 +1,201 @@
...
+.SILENT: kubeconfig.dec
+kubeconfig.dec: check-env-dir
+	[ ! -e $(ENV_DIR)/$(basename $(@)) ] && exit 0 \
+	|| ( \
+		sops -d $(ENV_DIR)/$(basename $(@)) > $(ENV_DIR)/$(@) || rm $(ENV_DIR)/$(@); \
+		test -s $(ENV_DIR)/$(@) || (echo "[ERR] Decryption failed: $(basename $(@))" && exit 1) \
+	)
+
+.PHONY: clean
+clean-decrypt: check-env-dir
```
jschaul, line 164: .PHONY and actual makefile target name don't match

jschaul (approved):
I didn't try this out, but on the surface it looks good. Found two typos 😸 .

Makefile
```diff
@@ -0,0 +1,201 @@
...
+.PHONY: re-init
+re-init: check-inputs-terraform
+	make --always-make $(TF_DATA_DIR)
+
+# NOTE: alias for 're-init'
+.PHONY: init
+.SILENT: init
+init:
+init:
+	echo "[WARN] 'make init' is deprecated. Please use re-init moving forward"
```
akshaymankar, line 43: Why keep deprecated things in a new file? Why not just delete this target?

akshaymankar (approved):
Looks good, I haven't run it either. I think the `init` target can be deleted, but I am not so sure.

Makefile
lucendio: I'm fine with that too. I tried to migrate the *interfaces* as they were and only get rid of the redundant stuff. I wanted to make changing CICD scripts a little bit less error-prone. I'll remove `make init`. We then just have to be a bit more conscious when making the changes there.

lucendio:
> I didn't try this out, but on the surface it looks good. Found two typos 😸 .

Great findings. Thanks a lot!

PR #448, opened 2021-04-19
[tooling] Cherry-pick a couple of targets and update documentation

lucendio:
This changeset adds a couple of targets from #432

* helmfile invocation (prefixed with 'hf-')
* ansible-play-% to invoke single playbooks from ./terraform/
* a target to override roles, specifically to install an older version of
  Kubespray (this didnt get relocated during initial effort of merging the
  Makefiles due to an async merge of #435)

Aside from that, this change set includes:

* simplify make code
* rename create-inventory to generate-inventory
  * documenting recent changes of the make interface of this repository

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,8 @@ $(TF_DATA_DIR):
 re-init: check-inputs-terraform
 	make --always-make $(TF_DATA_DIR)
 
-.PHONY: apply plan console destroy
-apply plan console destroy: check-inputs-terraform $(TF_DATA_DIR)
+.PHONY: apply plan console destroy refresh
+apply plan console destroy refresh: check-inputs-terraform $(TF_DATA_DIR)
 	cd $(TERRAFORM_WORKING_DIR) \
 	&& source $(ENV_DIR)/hcloud-token.dec \
 	&& terraform $(@) -var-file=$(ENV_DIR)/terraform.tfvars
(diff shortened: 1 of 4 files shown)


jmatsushita:
Hi @lucendio It looks good to me overall, my only remark would be that maybe using consistent abbreviated prefixes (like `hf-` as you already did) will help to make it more obvious what underlying tooling and therefore phase of the deployment (infra, k8s, apps) we're dealing with. I suppose we could also use `infra-`, `config-`, `app-` as prefixes. Basically anything that helps "self-document" and make things easier to reason about would be awesome. Pick what you like!

jmatsushita (approved):

jmatsushita:
> Should the ./[terraform,ansible,helm]/README.md merged into sth placed at the root of this repository?

Yes!

> Does it make sense to make the target generate-inventory a dependency of bootstrap and provision-sft instead of explicitly requiring it?

Good question I'm not sure. I think to answer this properly we'd want to think about our Day-1 vs Day-2 usage patterns?

lucendio:
> > Does it make sense to make the target generate-inventory a dependency of bootstrap and provision-sft instead of explicitly requiring it?
> 
> Good question I'm not sure. I think to answer this properly we'd want to think about our Day-1 vs Day-2 usage patterns?

Can you elaborate on the use cases or potential issues that you have in your head? 

I'm mostly concerned about the implicit  chain of dependency we would introduce? worst case when running for example the `bootstrap` target:
1. `terraform init`
2. `terraform output`
3. `ansible-playbook`

But on the other hand, it ensures, that the inventory - potentially changed due to a `terraform apply` that might have happened in between - is up to date be for a *play*.

jmatsushita:
> That is why I hesitated. Though, I think the advantaged for the user interface would outweigh the increased complexity. Thoughts?

I think it's worth the complexity cost, especially since you've nicely abstracted the pattern.

> I'm mostly concerned about the implicit chain of dependency we would introduce?

yes that's what I meant. On day 1 it makes sense to write the inventory. But on day 2, if you're playing with changes in the terraform and also the ansible or helm and want to apply a helm change not having committed your tf changes, with a dependency it would actually run the terraform changes (IIUC) which would be unexpected and possibly dangerous.

jmatsushita:
Since we'll be moving hopefully soon to running IaC commands in pipelines, maybe that's a non issue and we should be careful until then.

lucendio:
I could think of having two different `make` interfaces (not necessarily in two different files but defined by two different patterns). One used by machines (e.g. in automation), and the other used by humans during development. For the latter, an implicit `generate-inventory` would be fine IMHO. Make tagets that follow the general pattern outlined above, and another abstraction on top only for humans. Which would also allow to put in some fail-safes for the human interface.

PR #449, opened 2021-04-20
[ansible] Update SFT role to v0.4.1

lucendio:
To prevent nginx from crashing due to a type conversion issue.
See https://github.com/wireapp/ansible-sft/pull/32


lucendio:
merging this to fix SFT CD

PR #450, opened 2021-05-03
Remove dynamodb setting

arianvp:
Not needed anymore

### DIFF
--- a/values/wire-server/prod-values.example.yaml
+++ b/values/wire-server/prod-values.example.yaml
@@ -27,11 +27,14 @@ brig:
     elasticsearch:
       host: elasticsearch-external
     useSES: false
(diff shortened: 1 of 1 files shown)


jmatsushita:
Wouldn't we want to keep dynamoDB settings commented out as documentation if this is an example file?

arianvp:
that's what I am doing here no?

jmatsushita:
Indeed it is 😓

jmatsushita (approved):

arianvp:
The CI error is a known flake. GPG is my nemesis and I have no idea why sometimes it is convinced the private key isn't there. It happens once every 20 times on my computer as well 🤷🏾

jmatsushita:
Oh yeah, I saw that one before :/ Please go ahead and merge 👍

PR #452, opened 2021-05-06
Update wire-server version to 2.106.0

arianvp:


lucendio (approved):

arianvp:
Note to self: add a `gpg --list-private-keys` call before the GPG failure to further debug this issue in CI. Can't reproduce locally

PR #454, opened 2021-05-26
Release 2021-05-26

fisx:
See 74d6bdd


arianvp:
I think we don't need to do a wire-server-deploy release at this point.  Develop and master are 100% identical; except for  me trying to fix github actions today.

If some offline deploy customer is dependent on the `wire-server` release the following line should be touched:
https://github.com/wireapp/wire-server-deploy/blob/develop/offline/ci.sh#L99

otherwise i'd just wait for us to do a new wire-server-deploy release once there is a new k8s version

fisx:
cool, thanks for the context!

PR #456, opened 2021-05-28
Try add fluentbit and kibana

arianvp:
Lets see if it breaks


arianvp:
This needs rebasing once `2.110.0` is out

PR #474, opened 2021-06-15
allow custom private terraform snippets during the use of terraform e…

jschaul:
`…nvironment` usage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,15 @@ endif
 export TF_DATA_DIR = $(ENV_DIR)/.terraform
 TERRAFORM_WORKING_DIR = $(MKFILE_DIR)/terraform/environment
 
+# Since the variable directory and the terraform working directory are not the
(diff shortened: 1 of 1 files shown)


arianvp:
I think this demonstrates the smell I was a bit worried about and we broadly discussed a few months ago.

@jmatsushita this mostly seems to be a work around for the pattern that we only have `tfvars` in cailleach at the moment.  I think my suggestion to have only modules in `wire-server-deploy` and have actual terraform code in `cailleach` call those modules; instead of using just `.tfvars`  is probably a better way to tackle this. Thoughts? That way people can always add additional terraform code in cailleach if needed.


I'm ok with merging this for now; but I think the better way is to actually have the main terraform file in cailleach and have that call modules in `wire-server-deploy` so that we don't need these kind of workarounds but just let the tool do its job

arianvp (approved):

PR #476, opened 2021-06-23
Bump wire-server charts to 2.109.0

arianvp:


arianvp:
We'll need to wait for a `2.10.0` release. There is a known (not very critical) bug in team  settings (Doesn't display correct error message in some cases) that we want to have in the next offline release

PR #483, opened 2021-07-05
Fix sft ansible inventory to be compatible with newer kubespray

arianvp:


julialongtin (approved):
Tested in customer environment.

PR #486, opened 2021-07-06
actually fix nested values for prometheus-operator (cf. #420)

tjanson:
This time the components' config trees are actually correctly nested under `prometheus-operator` and their respective top levels (e.g. `prometheus`).

For reference, see the [values file](https://github.com/helm/charts/blob/9c2bd6d59c33ddca798e549f3030439465f3f094/stable/prometheus-operator/values.yaml#L1143-L1251) of `prometheus-operator` at the time.

Fix-up for #420.

### DIFF
--- a/values/wire-server-metrics/demo-values.example.yaml
+++ b/values/wire-server-metrics/demo-values.example.yaml
@@ -4,10 +4,11 @@
 #   grafana:
 #     persistence:
 #       enabled: false
-# prometheusSpec:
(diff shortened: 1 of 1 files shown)


tjanson:
This also needs to be fixed in the docs (not the indentation, but nesting `prometheusSpec` under `prometheus`).

jschaul:
Thank you!

jschaul (approved):

PR #490, opened 2021-07-16
Also add the other inventory's kubeconfigs

arianvp:
I'm not sure if this will work. I also remember Gregor made sure all the kubeconfigs were moved into the same location so maybe this isn't needed


julialongtin:
looks unnecessary. closing.

PR #491, opened 2021-08-05
Fixes needed to bootstrap a new environment

akshaymankar:
The commits are unrelated, so it might be better to rebase and merge.


jschaul (approved):
Looks reasonable. I think instead of the become:false here; it's also needed for kubespray, so the become=true should be removed from the ansible.cfg file instead perhaps.

PR #493, opened 2021-08-18
Add a method to monitor the progress of the tarball extraction

arthurwolf:
As during a training session the extraction was slow and we had no way to know how much time was remaining.


flokli:
I'm not sure we should document `pv` here. We can probably just add a `tar` invocation with the `v` argument instead of writing to extract the tarball literaly.

PR #494, opened 2021-08-19
Remove redis-ephemeral step

arthurwolf:
As it is already done in a previous step (listening during akamaya teaching session, I think I hear the step called "SNS"(?) but am not sure I heard correctly).


julialongtin (approved):
:perfect:

julialongtin (approved):

PR #501, opened 2021-09-13
Simplify SNS platform creation by splitting up in two resources

arianvp:
Instead of fighting Terraform by first differentiating into two resource types, to then merge them, and then split them up again;  just create a resource per application type.  The types then magically line up; and makes the code way easier to read in my opinion.

What I don't know is if this refactoring can be applied without downtime. But the reduction in complexity seems worth it.

Perhaps this needs some state surgery with `terraform state mv`  to move the respective apps into the `ios` and `android` resource


arianvp:
This doesn't work. still need to loop over platforms

PR #503, opened 2021-09-22
Adding logging playbook as per JCT-62

arthurwolf:
and matching templates


ansible/logging.yml
```diff
@@ -0,0 +1,36 @@
...
+      become: true
+      template:
+        src: templates/qradar.conf.j2
+        dest: /etc/rsyslog.d/rfc5424-remote.conf
+
+    # sudo chmod 775 /var/log/elasticsearch
+    - name: Change permissions of elastisearch log file
+      become: true
+      file:
+        path: /var/log/elasticsearch
```
arianvp, line 28: I'm a bit confused. this directory doesn't exist on all hosts. Is this only relevant for the elasticsearch role?

arthurwolf: fixed

PR #507, opened 2021-11-02
//terraform/modules/aws-gundeck-push-notifications: remove

flokli:
This isn't used anywhere outside of wire, and it explicitly targets push
credentials in the official wire product.


flokli:
GH Actions failure will be fixed by https://github.com/wireapp/wire-server-deploy/pull/508.

jschaul (approved):
I see this was refactored in cailleach, cool beans!

(I suppose in a hypothetical case of someone using a custom app build themselves and another means of distributing apps to their users, they might be interested in setting up SQS/SNS resources with their own certificates and apps and might therefore want to use this module; but; I admit the chances of this are very very slim, so, delete this code 😄 )

PR #514, opened 2021-11-22
nix: add nix-prefetch-docker to the env

flokli:
This is invoked by the `deps.yml` pipeline.


flokli:
Confirmed this works - manually triggered, it opened https://github.com/wireapp/wire-server-deploy/pull/515.

PR #526, opened 2021-12-13
Elasticsearch: set formatMsgNoLookups

jschaul:
See in-line comment.

### DIFF
--- a/ansible/elasticsearch.yml
+++ b/ansible/elasticsearch.yml
@@ -16,6 +16,9 @@
     es_xpack_features: [] # disable features
     es_instance_name: "{{ ansible_hostname }}"
 
+    # Protect against potential information leak.  
+    # For details see https://docs.wire.com/security-responses/log4shell.html
+    #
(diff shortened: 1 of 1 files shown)


ansible/elasticsearch.yml
```diff
@@ -16,6 +16,19 @@
     es_xpack_features: [] # disable features
     es_instance_name: "{{ ansible_hostname }}"
 
+    # Note: while elasticsearch stores some user provided data, it normally does
```
fisx, line 19:
```suggestion
    # Protect against potential information leak.  
    # For details see https://docs.wire.com/security-responses/log4shell.html
    #
    # Note: while elasticsearch stores some user provided data, it normally does
```

If you want to be extra nit-picky, you can move the entire comment block closer to line 30.

fisx (approved):

ansible/elasticsearch.yml
jschaul: thanks!

PR #528, opened 2021-12-15
update ES version and adjust comment

jschaul:
TBD: test this works.


julialongtin:
Looks like this worked.

julialongtin (approved):

PR #597, opened 2023-01-05
[DO NOT MERGE] add examples values for roman

amitsagtani97:
Example values for roman and postgresql13-ephemeral chart.

Depends on - https://github.com/wireapp/wire-server/pull/2935


values/roman/prod-values.example.yaml
```diff
@@ -0,0 +1,21 @@
+roman:
+  replicaCount: 1
+  image:
+    repository: quay.io/wire/roman
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: staging
```
julialongtin, line 6: do we really not have a production tag for this?

values/roman/prod-values.example.yaml
```diff
@@ -0,0 +1,21 @@
...
+  serviceToken: # Provide the roman service token.
+  # randomly generated for local testing
+  appKey: b53181dd-6400-4960-8988-f775545588ff-0949f503-421e-4588-a2c5-f64fd9c180fd
```
julialongtin, line 9: and how is the user supposed to randomly generate this?

values/roman/prod-values.example.yaml
```diff
@@ -0,0 +1,21 @@
...
+  wireApiHost: https://nginz-https.example.com
```
julialongtin, line 10: this is totally wrong. look at the prod-values.example.yaml . when referring to a wire component, we don't use FQDNs.

values/roman/prod-values.example.yaml
```diff
@@ -0,0 +1,21 @@
...
+  proxyDomain:
+  # random base 64 string used as certificate, replace with real one
+  pubKeyBase64: LS0tLS1CRUdJTiBQVcFQvWjcKMlczd1A1TzJveFVCTElDIEtFWS0tLS0tCg==
```
julialongtin, line 13: again, how is a user supposed to use this, and.. what kind of cert is this?

values/roman/prod-values.example.yaml
```diff
@@ -0,0 +1,21 @@
...
+  publicURL: roman.example.com # Link to the subdomain pointing to roman.
+  dbURL: jdbc:postgresql://postgresql:5432/externalPostgresql # Link to your Postgres server.
```
julialongtin, line 15: this should point to the ephemeral postgres, by default.

values/roman/prod-values.example.yaml
```diff
@@ -0,0 +1,21 @@
...
+  dbUser: externalPostgresql # Provide your Postgres DB username.
+  dbPassword: externalPostgresql # Provide your Postgres DB password.
+
+tls:
```
julialongtin, line 19: is there a way to supply certificates other than letsEncrypt?

julialongtin (changes requested):

PR #600, opened 2023-01-25
helm mapkubeapis plugin - also in offline packages

jschaul:
See also https://github.com/wireapp/wire-server/pull/3019 and https://github.com/zinfra/cailleach/pull/1502

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -1,4 +1,7 @@
-self: super: {
+self:
+let helm-mapkubeapis = self.callPackage ./pkgs/helm-mapkubeapis.nix { };
+in
+super: {
   pythonForAnsible = (self.python3.withPackages (_: self.ansible.requiredPythonModules ++ [
     super.python3Packages.boto
     super.python3Packages.boto3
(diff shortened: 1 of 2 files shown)


nix/overlay.nix
```diff
@@ -1,4 +1,7 @@
-self: super: {
+self:
+let helm-mapkubeapis = self.callPackage ./pkgs/helm-mapkubeapis.nix { };
```
supersven, line 2: I was a bit surprised seeing this expression here.

Have you tried to move it closer to its usage? E.g. to line 16?

supersven (approved):
Overall, LGTM :+1: 

I've left one comment. Solving it might make the code easier to digest.

nix/overlay.nix
jschaul: I did. But then it gave me "variable helm-mapkubeapis unknown" or something similar. When I moved it up here it worked.

supersven: It's not that bad. So, never mind :shrug: :smile:

PR #633, opened 2023-07-05
migrate to ingress-nginx-controller chart

amitsagtani97:


amitsagtani97:
this pr is no longer valid...as these changes needs to be merged along with migration to ubuntu 22...as this chart's minimum requirement is k8s 1.23

PR #647, opened 2023-09-07
Replace multiple sudo command with a single one

smatting:
Motivation: If you copy-paste a multiline string that has multiple sudo commands into a terminal not all commands will be executed and there will be no error to indicate that this happens. Since some people might copy multiple lines at once we should avoid this in our documentation.

Some of the conversions were made with this script

```python3
#!/usr/bin/env python3

import sys
import re

lines = sys.stdin.read().splitlines()
print('sudo bash -c \'')
print('set -eo pipefail;\n')
for line in lines:
    m = re.match('sudo (.*)', line)
    if m is None:
        sys.stderr('All lines have to start with sudo')
        sys.exit(1)
    cmd, = m.groups()
    if not cmd.endswith(';'):
        cmd += ';'
    print(cmd)
print('\'')
```

### DIFF
--- a/offline/ubuntu22.04_installation.md
+++ b/offline/ubuntu22.04_installation.md
@@ -215,11 +215,15 @@ tar -xf debs-jammy.tar
 ### (FIXME: add iptables to the repo) Install Docker from debian archive.
 
 ```
-sudo apt -y install iptables
-sudo dpkg -i debs-jammy/public/pool/main/d/docker-ce/docker-ce-cli_*.deb
-sudo dpkg -i debs-jammy/public/pool/main/c/containerd.io/containerd.io_*.deb
-sudo dpkg -i debs-jammy/public/pool/main/d/docker-ce/docker-ce_*.deb
-sudo dpkg --configure -a
+sudo bash -c '
+set -eo pipefail;
+
+apt -y install iptables;
+dpkg -i debs-jammy/public/pool/main/d/docker-ce/docker-ce-cli_*.deb;
+dpkg -i debs-jammy/public/pool/main/c/containerd.io/containerd.io_*.deb;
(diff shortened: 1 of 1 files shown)


offline/ubuntu22.04_installation.md
```diff
@@ -336,22 +370,32 @@ ip ro | sed -n "/default/s/.* dev \([en\(ps|o)0-9]*\) .*/export OUTBOUNDINTERFAC
...
-sudo sed -i 's/.*DEFAULT_FORWARD_POLICY=.*/DEFAULT_FORWARD_POLICY="ACCEPT"/' /etc/default/ufw
-sudo sed -i "1i *nat\n:POSTROUTING ACCEPT [0:0]\n-A POSTROUTING -s 172.16.0.0/24 -o $OUTBOUNDINTERFACE -j MASQUERADE\nCOMMIT" /etc/ufw/before.rules
-sudo service ufw restart
+sudo bash -c '
+set -eo pipefail;
+
+sed -i 's/.*DEFAULT_FORWARD_POLICY=.*/DEFAULT_FORWARD_POLICY="ACCEPT"/' /etc/default/ufw;
+sed -i "1i *nat\n:POSTROUTING ACCEPT [0:0]\n-A POSTROUTING -s 172.16.0.0/24 -o $OUTBOUNDINTERFACE -j MASQUERADE\nCOMMIT" /etc/ufw/before.rules;
+service ufw restart;
+'
```
smatting, line 379: ouch, that's not correct

smatting: Actually they get interpolated

smatting: No they don't. The whole block only works because there are unescaped `'`

PR #657, opened 2023-09-29
[DO NOT MERGE] Save point: Fix seeding containerd save

supersven:
This is just my save point for a big interactive rebasing session.

(**Just ignore it**)


supersven:
Overtaken by events.

PR #662, opened 2023-10-06
remove ubuntu 18 support

amitsagtani97:
Remove the support for ubuntu 18 based wire-server deployment in offline env.

#WPB-4880


julialongtin:
This removes the mechanics, but does not remove any of the documentation. can we verify that everything of import from the old documentation has been merged, and then remove it with this PR, as well?

supersven:
> This removes the mechanics, but does not remove any of the documentation. can we verify that everything of import from the old documentation has been merged, and then remove it with this PR, as well?

Adding to this: IMHO it would also be good to rename the existing `offline/ubuntu22.04_installation.md` to communicate that it is about a very specific setup; namely single Hetzner machine with KVM. (I like the name `single_hetzner_machine_installation.md`, but others may be even better.)

amitsagtani97:
Moved the restund firewall fixes docs and renamed the file to single_hetzner_machine_installation.md :)

julialongtin (approved):

arthurwolf:
+1

PR #666, opened 2023-10-11
Fix broken CI

amitsagtani97:
Fix the broken CI

### DIFF
--- a/offline/cd.sh
+++ b/offline/cd.sh
@@ -7,9 +7,9 @@ function cleanup {
   echo done
 }
 trap cleanup EXIT
-
 (cd terraform/examples/wire-server-deploy-offline-hetzner ; terraform init ; terraform apply -auto-approve )
 adminhost=$(cd terraform/examples/wire-server-deploy-offline-hetzner ; terraform output adminhost)
+adminhost="${adminhost//\"/}" # remove extra quotes around the returned string
 ssh_private_key=$(cd terraform/examples/wire-server-deploy-offline-hetzner ; terraform output ssh_private_key)
 
 eval `ssh-agent`
--- a/default.nix
+++ b/default.nix
@@ -43,7 +43,7 @@ rec {
       moreutils
       skopeo
       sops
-      terraform_0_13
+      terraform_1
       yq
       create-container-dump
       list-helm-containers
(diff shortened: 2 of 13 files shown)


amitsagtani97:
@julialongtin @supersven The last green CI was found to be working with v1.43.0 of hcloud module(https://registry.terraform.io/providers/hetznercloud/hcloud/latest)
the most recent version 1.44 onwards have a lot of changes which are not supported by many existing modules in our code.

Also, I tried to play around with different versions of terraform(0.14 and 1.0),  but it didn't worked. 
Looks like this is the good fix at the current set of dependencies(both internaland external like kubespray)

offline/cd.sh
```diff
@@ -4,12 +4,11 @@ set -euo pipefail
 
 function cleanup {
   (cd terraform/examples/wire-server-deploy-offline-hetzner ; terraform destroy -auto-approve)
-  echo done
 }
 trap cleanup EXIT
-
 (cd terraform/examples/wire-server-deploy-offline-hetzner ; terraform init ; terraform apply -auto-approve )
 adminhost=$(cd terraform/examples/wire-server-deploy-offline-hetzner ; terraform output adminhost)
+adminhost="${adminhost//\"/}"
```
supersven, line 11: Why is this needed? Has the output format of `terraform output` changed between versions (`0.13` -> `1.x`)?

supersven (changes requested):
Thanks a lot for investigating in this!

- I think it would be good to cleanup the commit history (squash / drop / ...)
- We should have an answer for this: https://github.com/wireapp/wire-server-deploy/pull/666/files#r1357993681
- Are you sure that all Terraform usages still work? Or, does this PR need additional manual testing?

amitsagtani97:
> Thanks a lot for investigating in this!
> 
> * I think it would be good to cleanup the commit history (squash / drop / ...)
> * We should have an answer for this: https://github.com/wireapp/wire-server-deploy/pull/666/files#r1357993681
> * Are you sure that all Terraform usages still work? Or, does this PR need additional manual testing?



> Thanks a lot for investigating in this!
> 
> * I think it would be good to cleanup the commit history (squash / drop / ...)
> * We should have an answer for this: https://github.com/wireapp/wire-server-deploy/pull/666/files#r1357993681
> * Are you sure that all Terraform usages still work? Or, does this PR need additional manual testing?

 - Fixed the history.
 - There were extra double quotes returned from the terraform output from the new version, so had to remove those.
 - As of now, I tried to run "terraform init" command in all the environment directory in the wire-server-deploy and it returned no warnings or error message.

PR #675, opened 2024-01-16
Bump kubernetes to v1.28.2

amitsagtani97:
wpb-2324

This pr adds dependencies for deploying kubernetes v1.28.2 by default, with latest kubespray version.


bin/offline-cluster.sh
```diff
@@ -47,9 +47,7 @@ ansible-playbook -i $INVENTORY_FILE $ANSIBLE_DIR/seed-offline-containerd.yml
 ansible-playbook -i $INVENTORY_FILE $ANSIBLE_DIR/sync_time.yml -v
 
 # Run the rest of kubespray. This should bootstrap a kubernetes cluster successfully:
-ansible-playbook -i $INVENTORY_FILE $ANSIBLE_DIR/kubernetes.yml --skip-tags bootstrap-os,preinstall,container-engine
-
-./bin/fix_default_router.sh
+ansible-playbook -i $INVENTORY_FILE $ANSIBLE_DIR/kubernetes.yml --skip-tags bootstrap-os,preinstall,container-engine,multus
```
julialongtin, line 50: multius?

offline/ci.sh
```diff
@@ -101,16 +116,16 @@ calling_charts=(
...
 # same as prior.. in most cases.
-wire_calling_version="4.39.0"
+wire_calling_version="4.39.159"
 
 # TODO: Awaiting some fixes in wire-server regarding tagless images
 HELM_HOME=$(mktemp -d)
 export HELM_HOME
 
-helm repo add wire https://s3-eu-west-1.amazonaws.com/public.wire.com/charts
+helm repo add wire https://s3-eu-west-1.amazonaws.com/public.wire.com/charts-develop
```
julialongtin, line 128: aha. this is smart. rewrite this file a bit, so this is the way we do this now?

julialongtin (approved):

bin/offline-cluster.sh
supersven: @julialongtin I'm surprised as you are, but there are traces of *multus* in our code base: https://github.com/kubernetes-sigs/kubespray/blob//docs/multus.md?plain=1#L1

Me needs a <scratches-head> emoji.

default.nix
```diff
@@ -26,8 +26,9 @@ rec {
   env = pkgs.buildEnv {
     name = "wire-server-deploy";
     paths = with pkgs; [
-      ansible_2_11
       pythonForAnsible
+      ansible_2_15
```
supersven, line 30: Small hint - though it's not very bad: Usually people try to keep the change set (amount of changed lines) small. I.e. I would have just replaced the `1` with a `5`.

But, this is easy to grasp and fine for now.

offline/ci.sh
```diff
@@ -101,16 +116,16 @@ calling_charts=(
 )
 
 # wire_version=$(helm show chart wire/wire-server | yq -r .version)
-wire_version="4.39.0"
+wire_version="4.39.159"
```
supersven, line 119: Hm, so the version that is currently on staging will be the version that our users should install? Was this an active choice by someone? :thinking: 

Or, will this branch not be merged to `master`?

bin/offline-cluster.sh
```diff
@@ -47,9 +47,7 @@ ansible-playbook -i $INVENTORY_FILE $ANSIBLE_DIR/seed-offline-containerd.yml
 ansible-playbook -i $INVENTORY_FILE $ANSIBLE_DIR/sync_time.yml -v
 
 # Run the rest of kubespray. This should bootstrap a kubernetes cluster successfully:
-ansible-playbook -i $INVENTORY_FILE $ANSIBLE_DIR/kubernetes.yml --skip-tags bootstrap-os,preinstall,container-engine
-
-./bin/fix_default_router.sh
```
supersven, line 52: I cannot find where this script is called now. Am I'm blind (probably) or is `fix_default_router.sh` now unused?

nix/pkgs/kubernetes-tools.nix
```diff
@@ -3,7 +3,7 @@
 
 buildGoModule rec {
   pname = "kubernetes";
-  version = "1.23.16";
+  version = "1.28.2";
```
supersven, line 6: I'm a bit surprised that the `hash` didn't change. Have you checked if this intented change became active?

On my machine there's still the old version in the nix env:

```
➜  wire-server-deploy git:(bump_kubernetes_to_1_28) kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.16", GitCommit:"60e5135f758b6e43d0523b3277e8d34b4ab3801f", GitTreeState:"archive", BuildDate:"1980-01-01T00:00:00Z", GoVersion:"go1.21.4", Compiler:"gc", Platform:"linux/amd64"}
```

nix/pkgs/wire-binaries.nix
```diff
@@ -8,14 +8,14 @@ let
   # These values are manually kept in sync with:
   # https://github.com/kubernetes-sigs/kubespray/blob/release-2.20/roles/download/defaults/main.yml
```
supersven, line 9: You seem to have updated `kubespray`. Is this comment still fine. Or, does it need to be updated, too?

offline/ci.sh
supersven: The clean solution would be to extract parameters for the parts that depend on the version. And, fill with with what we get from Stefan M.'s `build.json` files: https://github.com/wireapp/wire-builds/blob/108f7eb93bc6d1227f04c23032fab0c58390e5fe/build.json

This refactoring doesn't have to be done now. However, it's good to keep in mind that this changing of versions / Helm repos buried deep inside some files is neither the goal nor the future.

nix/pkgs/kubernetes-tools.nix
amitsagtani97: nice find.. I need to update the hash

nix/pkgs/wire-binaries.nix
amitsagtani97: needs to be updated..fixed :)

bin/offline-cluster.sh
amitsagtani97: this script is no longer needs to be executed, there was some issue with dns configs on k8s v1.23.7 which is no longer found in this newer version.

default.nix
amitsagtani97: makes sense, thanks for the suggestion :)
Was unintentional change, I appreciate this suggestion for my long term learning :)

offline/ci.sh
amitsagtani97: there were changes needed in the wire-server helm chart to support k8s 1.28
This change is temporary, goal is to replace it with "4.40" as soon as that is released and than merge

bin/offline-cluster.sh
amitsagtani97: this multus network plugin playbooks were being run while installing k8s, even the variable to install multus was set to false and was throwing some dependency issues, hence added that tag in the skip list

offline/ci.sh
amitsagtani97: this was temporary change to test the changes needed in helm charts for k8s 1.28
Goal is to use the build.json later.

PR #687, opened 2024-03-11
Wpb 6455 coturn docs

arthurwolf:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

_Briefly describe the issue you have solved or implemented with this pull request. If the PR contains multiple issues, use a bullet list._

### Causes (Optional)

_Briefly describe the causes behind the issues. This could be helpful to understand the adopted solutions behind some nasty bugs or complex issues._

### Solutions

_Briefly describe the solutions you have implemented for the issues explained above._

### Dependencies (Optional)

_If there are some other pull requests related to this one (e.g. new releases of frameworks), specify them here._

Needs releases with:

- [ ] GitHub link to other pull request

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

#### How to Test

_Briefly describe how this change was tested and if applicable the exact steps taken to verify that it works as expected._

### Notes (Optional)

_Specify here any other facts that you think are important for this issue._

### Attachments (Optional)

_Attachments like images, videos, etc. (drag and drop in the text box)_

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

### DIFF
--- a/offline/coturn.md
+++ b/offline/coturn.md
@@ -147,9 +147,44 @@ If you want to run Coturn on multiple machines, you must:
 
 ## Configure the SFT labels for Coturn and SFT to share a port range. 
 
+First, we must locate what the "external" IP address of the machine is. 
+
+We get it by running the following command:
+
+```bash
+
+sudo ip addr
+
+```
+
+The first interface will be the loopback interface, `lo`, and the second interface will be the "external" interface, `enp41s0` in our example, the output looking something like this:
+
+```bash
+
+demo@install-docs:~/wire-server-deploy$ ip addr
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
+    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host 
+       valid_lft forever preferred_lft forever
+2: enp41s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
+    link/ether a8:a1:59:a2:9b:5b brd ff:ff:ff:ff:ff:ff
+    inet 5.9.84.121/32 scope global enp41s0
+       valid_lft forever preferred_lft forever
+    inet6 2a01:4f8:162:3b6::2/64 scope global 
+       valid_lft forever preferred_lft forever
+    inet6 fe80::aaa1:59ff:fea2:9b5b/64 scope link 
+       valid_lft forever preferred_lft forever
+3: etc... 
+
+``` 
+
+In this case, the external IP address is `5.9.84.121`.
+
 ```{note}
 
-Note this step is also documented in the Switch to [ Wire install docs](docs_ubuntu_22.04.md)
+Note this step is also documented in the [ Wire install docs](docs_ubuntu_22.04.md)
 
 ``` 
 
@@ -178,6 +213,14 @@ d kubectl label node kubenode1 wire.com/role=sftd
 
 ```
 
+We must also annotate the node with the exrenal IP address we will be listening to (which we found with `sudo ip addr` above):
+
+```bash
+
+kubectl annotate node kubenode3 wire.com/external-ip='your.public.ip.address'
+
+```
+
 If we want to run SFT on multiple nodes, the procedure is the same as the one documented above for running Coturn on multiple nodes.
 
 We now should have Coturn configured to run on one or more kubernetes node(s), and SFT configured to run on one or more kubernetes node(s), and the two should not run on the same node(s)/overlap.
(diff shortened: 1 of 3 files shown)


offline/coturn.md
```diff
@@ -0,0 +1,471 @@
...
+
+# Value file for coturn chart.
+#
+# See: https://github.com/wireapp/wire-server/blob/develop/charts/coturn/values.yaml
+# And: https://github.com/wireapp/wire-server/blob/develop/charts/coturn/README.md
+
+nodeSelector:
+  wire.com/role: coturn
+
+coturnTurnListenIP: '192.168.122.23'
```
julialongtin, line 64: shouldn't we have a listen IP, and a publish IP?

offline/coturn.md
```diff
@@ -0,0 +1,471 @@
...
+
+Before moving on, we must also re-deploy SFT's chart to apply the new configuration:
+
+```bash
+
+d helm upgrade --install sftd ./charts/sftd --set 'nodeSelector.wire\.com/role=sftd' --values values/sftd/values.yaml
+
+```
+
+## Configure the port redirection in Nftables.
```
julialongtin, line 193: we should mark this section as only if you are running behind a nftables managed firewall.

offline/coturn.md
```diff
@@ -0,0 +1,471 @@
...
+
+We must configure the port redirection in Nftables to allow traffic to reach Coturn and SFT.
+
+Calling and TURN services (Coturn, Restund, SFT) require being reachable on a range of ports used to transmit the calling data.
+
+Because SFT and Coturn can not run on the same node, and can not share the same port range on the "outside" interface of the machine, we must configure the port redirection in Nftables to allow traffic to reach Coturn and SFT by splitting the ports between the two services.
```
julialongtin, line 199: I am not certain this is true. it's MORE true that they both want to use the same port range, so predicting which nodes are using what ports ahead of time requires dividing ahead of time.

offline/coturn.md
```diff
@@ -0,0 +1,471 @@
...
+    v1: []
+    v2:
+      - "turn:<IP of restund1>:80"
+      - "turn:<IP of restund2>:80"
+      - "turn:<IP of restund1>:80?transport=tcp"
+      - "turn:<IP of restund2>:80?transport=tcp"
+
+``` 
+
+Instead, we configure it to use the external IP addres we found above, and the Coturn port, `3478` (as seen above in the `nftables` configuration):
```
julialongtin, line 375: you have missed the part where we annotate each node with it's expected external IP.

julialongtin (changes requested):

offline/coturn.md
```diff
@@ -0,0 +1,471 @@
...
+
+We get it by running the following command:
+
+```bash
+
+sudo ip addr
+
+```
+
+The first interface will be the loopback interface, `lo`, and the second interface will be the "external" interface, `enp41s0` in our example, the output looking something like this:
```
marcoconti83, line 312: What is the intended audience of this guide? Shouldn't we assume that someone following the guide is familiar with IP addresses and can figure out the IP address of a machine, i.e. we don't need to spell `sudo ip addr` for them?

offline/coturn.md
```diff
@@ -0,0 +1,471 @@
...
+
+```bash
+
+d helm upgrade --install wire-server ./charts/wire-server --values values/wire-server/values.yaml
+
+```
+
+## Disable Restund.
+
+As we are no longer using Restund, we should now disable it entirely.
```
marcoconti83, line 395: For a production environment, we can't disable Restund before clients have migrated to Coturn by retrieving the freshly updated call configuration that instructs them to use the Coturn IPs. 
This happens at least once every 24h (i.e. the clients pull the configuration every 24h if they are open).

So the process would be:
- Deploy Coturn instances
- Change turnStatic call config to point to Coturn IPs
- Wait 48h (24h call config pull delay + wait for any up to 24h long already existing call  to be over)
- Disable Restund

offline/coturn.md
arthurwolf: annotate the node with the outside IP

arthurwolf: same for coturn as it is for sft

arthurwolf: fixed in latest commit.

offline/coturn.md
arthurwolf: fixed in latest commit:

```{note}

Note: This section is only relevant if you are running Wire-Server/Coturn/SFT behind a `nftables`-managed firewall.

```

offline/coturn.md
arthurwolf: fixed in latest commit.

We must also annotate the node with the exrenal IP address we will be listening to (which we found with `sudo ip addr` above):

```bash

kubectl annotate node kubenode3 wire.com/external-ip='your.public.ip.address'

```

offline/coturn.md
arthurwolf: @marcoconti83 the general guidance for docs is we should be on the side of explaining more rather than explaining less. most users are tech-savvy, but I've had to walk some through some very basic things (even more basic than this). I asked Julia and she confirmed in this case it's fine.

offline/coturn.md
arthurwolf: added an appendix at the bottom of the document explaining this. thanks.

PR #693, opened 2024-04-17
Dnsmasq fix

jschumacher-wire:
https://wearezeta.atlassian.net/browse/WPB-8758

DO NOT let dnsmasq listen on external interface, we don't want to turn our Hetzner boxes into public resolvers.

### DIFF
--- a/ansible/hetzner-single-deploy.yml
+++ b/ansible/hetzner-single-deploy.yml
@@ -1,8 +1,8 @@
 - hosts: all
   become: true
   vars:
-    artifact_hash: 736b8ea76de5e67b68d15e8d2a70063b240a5af5
-    ubuntu_version: 22.04.4
+    artifact_hash: 94bdc70a1e9864261543e3ec3ffcd9069b09bd17
+    ubuntu_version: 22.04.3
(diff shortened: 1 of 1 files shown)


ansible/hetzner-single-deploy.yml
```diff
@@ -134,6 +134,15 @@
     virt_net:
       name: wirebox
       autostart: yes
+  - name: disable dnsmasq for public interface
+    lineinfile:
+      path: /etc/dnsmasq.d/disable_public_resolv
+      line: "except-interface={{ ansible_default_ipv4.interface }}"
```
supersven, line 140: Hm, I'm confused: Wouldn't it be better to explicitly name the interface to listen on `interface=...`? :thinking: 

But, I have to admit that I'm absolutely no expert in this area.

supersven (changes requested):

ansible/hetzner-single-deploy.yml
jschumacher-wire: It turns out we don't need dnsmasq at all, libvirt even sets a config in "/etc/dnsmasq.d-available/libvirt-daemon" to disable listening on virbr0 interface.  I remember dnsmasq used to be a dependency for libvirt, but that was 7-8 years ago, and I only touched pure libvirt / KVM hosts again here at Wire. My bad!

This PR now removes dnsmasq altogether.

supersven (approved):
LGTM :+1:

PR #703, opened 2024-05-15
ldap-scim-bridge docs update and add bzip2 library

amitsagtani97:
WPB-7170

ldap-scim-bridge chart is now available in offline artifacts through wire-builds directly. Updating docs for the same.

### DIFF
--- a/offline/ldap-scim-bridge.md
+++ b/offline/ldap-scim-bridge.md
@@ -1,11 +1,5 @@
 # How to deploy the ldap-scim-bridge
 
-Note: the LDAP Scim bridge is in a separate package at the moment. the docker container is available from: 
-
-https://temp-rhc-jun.s3.eu-west-1.amazonaws.com/ldap-scim-bridge%3A0.4.tar.bz2
-
-the helm chart is in wire-server.
-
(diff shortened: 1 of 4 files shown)


offline/ldap-scim-bridge.md
```diff
@@ -45,16 +39,6 @@ config:
...
 
-When you get the package:
-
-Copy your values and charts folders into the `Wire-Server` directory you're using.
-
-Copy the container image to all of the kubernetes hosts in your cluster.
-
-Pre-seed the docker container for `ldap-scim-bridge` onto all of your kubernetes hosts.
-```
-sudo bash -c "cat ldap-scim-bridge:0.4.tar.bz2 | docker load"
```
julialongtin, line 56: we are using containerd now. you also might want to show "copy the file" "ssh in" and "load the container" as separate steps.

amitsagtani97: Image is now downloaded already in the artifact and will be loaded into kubenodes as part of seed-offline-containerd.yaml

julialongtin (approved):

PR #710, opened 2024-06-10
bring back dnsmasq to vms

amitsagtani97:
hetzner-single-deploy playbook fails to setup WIAB env due to missing dnsmasq on host machine. Adding it back

### DIFF
--- a/ansible/hetzner-single-deploy.yml
+++ b/ansible/hetzner-single-deploy.yml
@@ -103,14 +103,11 @@
       owner: root
       group: root
     notify: sshd | restart
-  - name: Stop dnsmasq service
-    ansible.builtin.service:
+  - name: stop and disable dnsmasq service
+    service:
       name: dnsmasq
       state: stopped
-  - name: Disable dnsmasq service
-    ansible.builtin.service:
-      name: dnsmasq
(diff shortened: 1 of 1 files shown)


jschumacher-wire:
Thanks for the fix. Can you also add a task to stop and disable the dnsmasq service? The libvirt tasks in the playbook expect the binary to be there, but we don't want to accidentally create open resolvers again.

ansible/hetzner-single-deploy.yml
```diff
@@ -102,6 +103,14 @@
...
       group: root
     notify: sshd | restart
+  - name: Stop dnsmasq service
+    ansible.builtin.service:
+      name: dnsmasq
+      state: stopped
+  - name: Disable dnsmasq service
+    ansible.builtin.service:
+      name: dnsmasq
+      enabled: no
```
jschumacher-wire, line 113:
```suggestion
  - name: stop and disable dnsmasq service
    service:
      name: dnsmasq
      state: stopped
      enabled: false
```

jschumacher-wire (changes requested):

jschumacher-wire (approved):

PR #722, opened 2024-07-30
add keycloak, k8ssandra-operator and openebs helm chart

amitsagtani97:
https://wearezeta.atlassian.net/browse/WPB-10358


values/coturn/prod-values.example.yaml
```diff
@@ -0,0 +1,4 @@
+secrets:
+  zrestSecrets:
+    - ""
+  zrestSecret: ""
```
lwille, line 4: Are we sure this is needed? I couldn't find a reference in the `coturn` chart.

**suggestion:** delete.
```suggestion
```

values/restund/prod-values.example.yaml
```diff
@@ -0,0 +1,4 @@
+secrets:
+  zrestSecrets:
+    - ""
```
lwille, line 3: it seems `restund` only uses `zrestSecret`, so we can delete the other key from here.
```suggestion
```

lwille (changes requested):

PR #725, opened 2024-08-16
add sftTokenSecret reference

amitsagtani97:
https://wearezeta.atlassian.net/browse/WPB-10302

### DIFF
--- a/bin/offline-secrets.sh
+++ b/bin/offline-secrets.sh
@@ -39,6 +39,8 @@ brig:
     setNexmo: |-
       key: "dummy"
       secret: "dummy"
+    # Uncomment and update value to enable sft credentials provisioning
+    # sftTokenSecret: dummysecret
 cargohold:
   secrets:
     awsKeyId: "$minio_access_key"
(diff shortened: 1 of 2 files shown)


bin/offline-secrets.sh
```diff
@@ -39,6 +39,8 @@ brig:
     setNexmo: |-
       key: "dummy"
       secret: "dummy"
+    # Uncomment and update value to enable sft credentials provisioning
```
smatting, line 42: "Provisioning" is not correct, nothing is being provisioned

Maybe something like
```suggestion
    # Uncomment to connect to SFT servers that require it```

File: values/wire-server/prod-secrets.example.yaml [hunk 2]
```diff
@@ -22,8 +22,10 @@ brig:
     setNexmo: |-
       key: "dummy"
       secret: "dummy"
-    # Uncomment for federation. Set values accordingly
+    # Uncomment and update value to enable sft credentials provisioning
+    # sftTokenSecret: dummysecret
```
smatting, line 26: Same comment as above

values/wire-server/prod-secrets.example.yaml
```diff
@@ -22,8 +22,10 @@ brig:
...
 
+    # Uncomment for federation. Set values accordingly
```
smatting, line 28: > Set values accordingly
Sounds to me like this sentence is saying nothing

PR #728, opened 2024-08-30
Wiab automation

jschumacher-wire:
https://wearezeta.atlassian.net/browse/WPB-9381

### DIFF
--- a/bin/autodeploy.sh
+++ b/bin/autodeploy.sh
@@ -119,6 +119,7 @@ system_cleanup_meta() {
   msg ""
   sleep 5
   ssh -p "$SSH_PORT" "$SSH_USER"@webapp."$TARGET_SYSTEM" "bash -s" <<EOT
+# Making relevant vars and functions available to remote shell via SSH
 $(declare -p DEMO_USER)
 $(declare -f system_cleanup)
 system_cleanup
@@ -406,6 +407,7 @@ fi
 msg "INFO: Commencing Wire-in-a-box deployment on $TARGET_SYSTEM."
 preprovision_hetzner
 ssh -p "$SSH_PORT" "$DEMO_USER"@webapp."$TARGET_SYSTEM" "bash -s" <<EOT
+# Making relevant vars and functions available to remote shell via SSH
 $(declare -p DEMO_USER TARGET_SYSTEM SCRIPT_DIR)
 $(declare -f remote_deployment)
 remote_deployment


bin/autodeploy.sh
```diff
@@ -0,0 +1,416 @@
...
+  d helm install ingress-nginx-controller ./charts/ingress-nginx-controller --values ./values/ingress-nginx-controller/values.yaml
+
+  KUBENODEIP=$(d kubectl get pods -l app.kubernetes.io/name=ingress-nginx -o=custom-columns=IP:.status.hostIP --no-headers)
+  sudo sed -i "s/define KUBENODEIP.*/define KUBENODEIP = $KUBENODEIP/" /etc/nftables.conf
+  sudo systemctl restart nftables
+
+  INGRESSNODE=$(d kubectl get pods -l app.kubernetes.io/name=ingress-nginx -o=custom-columns=NODE:.spec.nodeName --no-headers)
+  d kubectl cordon "$INGRESSNODE"
+
+  wget https://charts.jetstack.io/charts/cert-manager-v1.13.2.tgz
```
lwille, line 320: is there a reason we manually download and unpack charts instead of using `helm repo add`?

bin/autodeploy.sh
```diff
@@ -0,0 +1,416 @@
...
+ansnode2
+ansnode3
+
+[rmq-cluster]
+ansnode1
+ansnode2
+ansnode3
+EOF
+
+  d ./bin/offline-cluster.sh
```
lwille, line 256: possibly dumb question, but where does `d` come from? It's hard to find a single-letter symbol.

bin/autodeploy.sh
```diff
@@ -0,0 +1,416 @@
...
+  msg ""
+  msg "WARNING: existing Docker containers found."
+  DO_SYSTEM_CLEANUP=true
+fi
+
+if [ "$DO_SYSTEM_CLEANUP" = false ]; then
+  msg ""
+  msg "INFO: Target system clean, no previous wire-server-deploy installation found."
+fi
+if [ "$DO_SYSTEM_CLEANUP" = true ] && [ "$FORCE_REDEPLOY" = 0 ]; then
```
lwille, line 392: **thought:** is there a reason to avoid an `else` branch and instead duplicate the conditions in the following `if`s?

```suggestion
elif [ "$FORCE_REDEPLOY" = 0 ]; then
```

bin/autodeploy.sh
```diff
@@ -0,0 +1,416 @@
...
+EOF
+
+  d ./bin/offline-cluster.sh
+  d kubectl get nodes -owide
+  ANSNODES="ansnode1 ansnode2 ansnode3"
+  for ANSNODE in $ANSNODES; do ssh -o StrictHostKeyChecking=no "$ANSNODE" "sudo bash -c '
+set -eo pipefail;
+
+#cassandra
+ufw allow 9042/tcp;
```
lwille, line 263: why do we use Ansible at all if we use imperative provisioning everywhere?

bin/autodeploy.sh
jschumacher-wire: Taken from https://github.com/wireapp/wire-server-deploy/blob/master/offline/docs_ubuntu_22.04.md

I guess the original thought was to avoid creating external dependencies for a "true" offline install, without internet. Which does not work anyway with (automated) demo WIAB, but I wanted to avoid upstreaming changes to the "official" offline documentation.

bin/autodeploy.sh
jschumacher-wire: The existing playbooks are supposed to take care of that AFAIK, which did not work reliably in the past - some deployments did have the rules applied, some did not.
I admit it's hacky, but it works reliably, and actual customer site deployments use their own networking / firewalling.

bin/autodeploy.sh
jschumacher-wire: The existing logic accounts for the case of a "clean" target system where no resource removal is necessary, in which we don't need to jump into the system_cleanup() function. Otherwise the output looks like this:

```
INFO: Target system clean, no previous wire-server-deploy installation found.

INFO: Cleaning up all VMs, docker resources and wire-server-deploy files on wiab-autodeploy.wire.link.
```

It's just cosmetics, but I's like to have consistent script output.

bin/autodeploy.sh
amitsagtani97: I think we can remove this manual pulling of cert-manager helm chart in another pr and use our new helm-charts / wire-builds repo to put in the offline artifact.

amitsagtani97:
While trying to run it locally for an installation, got the following error - 
```
INFO: running local ansible playbook for inital server deployment.
INFO: This will setup up the Hetzner system with basic defaults, download and unpack the wire-server-deploy artifact.
/nix/store/x1xxsh1gp6y389hyl40a0i74dkxiprl7-bash-5.2-p15/bin/bash: warning: setlocale: LC_ALL: cannot change locale (C.UTF-8): No such file or directory
ERROR: Ansible could not initialize the preferred locale: unsupported locale setting
```

bin/autodeploy.sh
```diff
@@ -0,0 +1,416 @@
...
+ * artifact-hash = 5c06158547bc57846eadaa2be5c813ec43be9b59
+ * target-system = wiab-autodeploy.wire.link
+
+Available options:
+-h, --help          Print this help and exit
+-v, --verbose       Print script debug info
+--force-redeploy    Force cleanup of previous Wire-in-a-box installation on target host
+--artifact-hash     String, specifies artifact ID as created here: https://github.com/wireapp/wire-server-deploy/actions/workflows/offline.yml
+                    Defaults to 5c06158547bc57846eadaa2be5c813ec43be9b59
+--target-system     String, domain name used to access the target host
```
arthurwolf, line 39: I think it would be good if this parameter contained the "domain" string somewhere, I initially expected it was like the OS version.

jschumacher-wire: Good point, I've renamed the `--target-system` flag to `--target-domain`.

amitsagtani97 (approved):

PR #734, opened 2024-09-16
add step-certificates example values

amitsagtani97:
https://wearezeta.atlassian.net/browse/WPB-11121


lwille (commented):
hack(smallstep-ca): drop unused connection test

related to WPB-11121

lwille:
currently, `docker pull docker://cr.k8ssandra.io/k8ssandra/cass-management-api:3.11.16` is returning a `500` error.
We should setup a transparent caching proxy (e.g. ECR) to mitigate such issues.

PR #742, opened 2024-10-24
Fix wiab automation

amitsagtani97:
https://wearezeta.atlassian.net/browse/WPB-11340

Fix the autodeploy script for installing wire-server on fresh vms

### DIFF
--- a/.github/workflows/deploy-wiab.yml
+++ b/.github/workflows/deploy-wiab.yml
@@ -1,4 +1,4 @@
-name: Deploy on Hetnzer WIAB setup
+name: Deploy on Hetzner WIAB setup
 on:
   pull_request:
     branches:
@@ -30,7 +30,7 @@ jobs:
       - name: Get latest commit SHA
         id: get_commit_sha
         run: |
-          COMMIT_SHA=$(git rev-parse --short HEAD)
+          COMMIT_SHA=$(git rev-parse HEAD)
           echo "commit_sha=$COMMIT_SHA" >> $GITHUB_ENV
 
       # Step 4: Run the autodeploy script
--- a/.github/workflows/deploy-wiab.yml
+++ b/.github/workflows/deploy-wiab.yml
@@ -1,8 +1,5 @@
 name: Deploy on Hetzner WIAB setup
 on:
-  pull_request:
-    branches:
-      - master
   workflow_run:
     workflows: ["Prepare custom offline package"]
     types:
--- a/.github/workflows/deploy-wiab.yml
+++ b/.github/workflows/deploy-wiab.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up SSH key
         uses: webfactory/ssh-agent@v0.5.3
         with:
-          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+          ssh-private-key: ${{ secrets.WIAB_PRIVATE_SSH_KEY }}
     
       # Step 3: Get the latest commit SHA, for the artifact
       - name: Get latest commit SHA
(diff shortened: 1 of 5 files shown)


bin/autodeploy.sh
```diff
@@ -134,12 +136,22 @@ system_cleanup() {
   rm -f /home/$DEMO_USER/wire-server-deploy-static-*.tgz
 }
 
+# Make sure the host machine is up to date and has all required packages installed
+ssh -p "$SSH_PORT" "$SSH_USER"@webapp."$TARGET_SYSTEM" "
+  sudo apt update && sudo apt upgrade -y && sudo apt install -y aptitude apt-transport-https bind9-host curl debian-goodies dnsutils git dnsmasq less lsof net-tools rsyslog screen sudo vim wget whois docker.io telnet python3-lxml qemu qemu-kvm qemu-utils libvirt-clients libvirt-daemon-system virtinst bridge-utils
```
jschumacher-wire, line 141: I think we should let the Hetzner ansible playbook take care this.

You want to make sure that the `system_cleanup()` function does not fail when trying to run the script on a clean host, right?

I would create an `if` branch within that function to determine if the commands are available. Something like this:

```
system_cleanup() {
  if which docker > /dev/null && which virsh > /dev/null
  then
    for VM in $(virsh list --all --name); do virsh destroy "$VM"; virsh undefine "$VM" --remove-all-storage; done
    docker system prune -a -f
  fi
  rm -f /home/$DEMO_USER/.ssh/known_hosts
  rm -rf /home/$DEMO_USER/wire-server-deploy
  rm -f /home/$DEMO_USER/wire-server-deploy-static-*.tgz
}
```

jschumacher-wire (changes requested):
Hi Amit, thank you for the fixes! I left a small suggestion.

bin/autodeploy.sh
amitsagtani97: @jschumacher-wire I have updated the script to handle more cases, PTAL

.github/workflows/deploy-wiab.yml
```diff
@@ -0,0 +1,40 @@
+name: Deploy on Hetnzer WIAB setup
```
jschumacher-wire, line 1:
```suggestion
name: Deploy on Hetzner WIAB setup
```

jschumacher-wire (changes requested):

jschumacher-wire:
I've added a small check in `bin/offline-vm-setup.sh` to wait and only exit the script once VM deployment is done. Otherwise, we might kill the small webserver used by cloud-init too early, potentially resulting in failed bootstrap installs for the last few VMs on the list.

Side note - I remember there was discussion regarding where to implement the pipeline, and that Concourse might be preferred. I realize it's not in the ticket, but I wanted to double-check.

@lwille what do you think?

In any case, very good work @amitsagtani97 :)

.github/workflows/deploy-wiab.yml
```diff
@@ -0,0 +1,37 @@
...
+        id: get_commit_sha
+        run: |
+          COMMIT_SHA=$(git rev-parse HEAD)
+          echo "commit_sha=$COMMIT_SHA" >> $GITHUB_ENV
+
+      # Step 4: Run the autodeploy script
+      - name: Run Auto Deploy Script
+        run: |
+          cd bin
+          ./autodeploy.sh --artifact-hash 2e8e4319ff3a8c309dc0ef5a1800611faa1ab0cb --target-domain wiab-test-box.wire.link --force-redeploy
```
jschumacher-wire, line 37: Shouldn't the `artifact-hash` ID be invoked dynamically from `$COMMIT_SHA` above?

amitsagtani97: nice catch, thanks. I forgot to remove the hard coded sha used for testing. Fixed now

jschumacher-wire (approved):

PR #765, opened 2025-03-04
remove terraform/ dir

smatting:
In cailleach this no longer referenced as of https://github.com/zinfra/cailleach/pull/2580

If there are no no other uses we can remove it here


smatting:
it's used by ci

PR #787, opened 2025-06-18
Create demo-values.example.yaml for postgresql chart

mohitrajain:


mohitrajain:
closing it since, this helm chart needs to be tested for build and deployment in prod and demo environments

PR #797, opened 2025-07-30
feat(postgres): WPB-17218 add ansible playbook to setup postgres HA

mohitrajain:
…luster

### Change type

* [ ] Fix
* [x] Feature
* [ ] Documentation
* [ ] Security / Upgrade

### Basic information 

* [ ] THIS CHANGE REQUIRES A DEPLOYMENT PACKAGE RELEASE
* [ ] THIS CHANGE REQUIRES A WIRE-DOCS RELEASE

### Testing

* [x] I ran/applied the changes myself, in a test environment.
* [ ] The CI job attached to this repo will test it for me.

### Tracking

* [ ] I added a new entry in an appropriate subdirectory of `changelog.d`
* [x] I mentioned this PR in Jira, OR I mentioned the Jira ticket in this PR.
* [ ] I mentioned this PR in one of the issues attached to one of our repositories.

### Knowledge Transfer
* [ ] An Asciinema session is attached to the Jira ticket.


bin/offline-helm.sh
```diff
@@ -23,7 +24,12 @@ helm upgrade --install --wait postgresql ./charts/postgresql --values ./values/p
 helm upgrade --install --wait rabbitmq ./charts/rabbitmq --values ./values/rabbitmq/prod-values.example.yaml --values ./values/rabbitmq/prod-secrets.example.yaml
 helm upgrade --install --wait databases-ephemeral ./charts/databases-ephemeral --values ./values/databases-ephemeral/prod-values.example.yaml
 helm upgrade --install --wait reaper ./charts/reaper
-helm upgrade --install --wait --timeout=30m0s wire-server ./charts/wire-server --values ./values/wire-server/prod-values.example.yaml --values ./values/wire-server/secrets.yaml
+helm upgrade --install --wait --timeout=10m0s wire-server ./charts/wire-server --values ./values/wire-server/prod-values.example.yaml --values ./values/wire-server/secrets.yaml || true
```
Veki301: I think timeout of 10m here will not be enough

ansible/helm_external.yml
```diff
@@ -46,3 +46,27 @@
         server_type: rmq-cluster
         network_interface: "{{ rabbitmq_network_interface }}"
   tags: rabbitmq-external
+
+- hosts: "postgresql"
```
Veki301, line 50: this entire thing seems it could have been more inline with the current approach and "template" we have in `tasks/helm-external.yml` maybe it can be structured in a similar fashion ? (I understand theres a caveat here with RWIPs and ROIPs, but maybe look into it?)

ansible/inventory/offline/group_vars/all/offline.yml
```diff
@@ -71,4 +72,37 @@ ingress_nginx_version: "v1.12.1"
 metrics_server_version: "v0.7.2"
 ingress_nginx_kube_webhook_certgen_image_tag: "v20231011-8b53cabe0"
 cert_manager_version: "v1.16.3"
-nginx_image_tag: "1.25.4-alpine"
\ No newline at end of file
+nginx_image_tag: "1.25.4-alpine"
+
+# postgresql configuration
```
Veki301: keep OVERRIDES at the bottom of the file, so theres no confusion wether or not postgres is a part of overrides or not

ansible/helm_external.yml
```diff
@@ -46,3 +46,27 @@
...
+  become: false
+  tasks:
```
Veki301, line 52: create a tag, we might wanna skip it if k8s postgres is being used (like we do for rmq)

ansible/postgresql-playbooks/postgresql-deploy-primary.yml
```diff
@@ -0,0 +1,92 @@
+- name: Deploy PostgreSQL Primary node
+  hosts: postgresql_rw
+  become: yes
+  gather_facts: yes
+  vars:
+    primary_node: "{{ hostvars[(groups.get('postgresql_rw', []) | first) | default('postgresql1')]['ansible_default_ipv4']['address'] | default(hostvars[(groups.get('postgresql_rw', []) | first) | default('postgresql1')]['ansible_host'] | default((groups.get('postgresql_rw', []) | first) | default('postgresql1'))) }}"
```
Veki301, line 6: jesus :D can this and any other similar example in the playbooks be made more readable/configurable through hosts.ini?

offline/tasks/pre_chart_process_0.sh
```diff
@@ -18,3 +18,5 @@ sed -i -Ee 's/useSharedFederatorSecret: false/useSharedFederatorSecret: true/' "
 # drop step-certificates/.../test-connection.yaml because it lacks an image tag
 # cf. https://github.com/smallstep/helm-charts/pull/196/files
 rm -v "${OUTPUT_DIR}"/charts/step-certificates/charts/step-certificates/templates/tests/*
+rm -v "${OUTPUT_DIR}"/charts/smtp/charts/smtp/templates/tests/*
```
Veki301: this removal of `smtp/templates/tests` should no longer be required

ansible/postgresql-playbooks/postgresql-deploy-replica.yml
```diff
@@ -0,0 +1,108 @@
+---
+- name: Deploy PostgreSQL replica services with streaming replication
+  hosts: postgresql_ro
+  become: yes
+  gather_facts: yes
+  vars:
+    primary_node: "{{ hostvars[(groups.get('postgresql_rw', []) | first) | default('postgresql1')]['ansible_default_ipv4']['address'] | default(hostvars[(groups.get('postgresql_rw', []) | first) | default('postgresql1')]['ansible_host'] | default((groups.get('postgresql_rw', []) | first) | default('postgresql1'))) }}"
```
Veki301, line 7: same comment for readability

ansible/postgresql-deploy.yml
```diff
@@ -0,0 +1,19 @@
+- name: Install PostgreSQL packages
+  import_playbook: postgresql-playbooks/postgresql-install.yml
```
Veki301, line 2: just for discussion, should the postgres playbook be in roles, and then the role invoked here ? @julialongtin

offline/postgresql-cluster.md
```diff
@@ -0,0 +1,483 @@
...
+- [Running the Playbook](#running-the-playbook)
+- [PostgreSQL Packages Installation Playbook](#postgresql-packages-installation-playbook)
+- [Deployment Architecture](#deployment-architecture)
+- [Monitoring and Verification](#monitoring-and-verification)
+- [Wire Server Database Setup](#wire-server-database-setup)
+- [Troubleshooting](#troubleshooting)
+- [Best Practices](#best-practices)
+- [Security Considerations](#security-considerations)
+
+## OverviewostgreSQL Cluster Deployment
```
Veki301: OverviewostgreSQL :)

ansible/postgresql-deploy.yml
```diff
@@ -0,0 +1,19 @@
...
+  when: skip_postgresql_install | default(false) == false
```
Veki301: we usually tag the playbooks and skip them with --skip-tags

offline/postgresql-cluster.md
```diff
@@ -0,0 +1,483 @@
...
+   - **Checksum Validation**: Ensures package integrity during download
+   - **Dependency Resolution**: Handles package dependencies automatically
+   - **Installation Verification**: Confirms successful installation of all components
+
+### Usage
+To run the [`postgresql-install.yml`](../ansible/postgresql-playbooks/postgresql-install.yml) playbook independently, use the following command:
+
+```bash
+ansible-playbook -i ansible/inventory/offline/hosts.ini ansible/postgresql-playbooks/postgresql-install.yml
+```
```
Veki301, line 248: another good example why tagging might be better, so user can be blissfully unaware where a particular playbook is located

bin/offline-helm.sh
mohitrajain: It was failing to deploy and was failing for 30mins, i had implemented a fast fail to find the errors. Now, it is able to get deployed within 10 mins but since it is a general script, i will increase it to 30mins.

ansible/helm_external.yml
mohitrajain: Yes for the reason of having two types of endpoints RW and RO, i had to change the templating and couldn't use helm-external.yml, hence created a specific task for postgresql

ansible/inventory/offline/group_vars/all/offline.yml
mohitrajain: sorry didn't follow. These vars are put here to make them available for `all` group. As they were new entry so I appended them here.

ansible/postgresql-deploy.yml
mohitrajain: yes, we could do that but here, all playbooks under `postgresql-playbooks` are specific to wire architecture, like 3 node config i.e. 1 primary and 2 replicas. Wire specific database and user creation. These are simple playbooks designed to do specific tasks, I am not sure, if they can be invoked as roles.

ansible/postgresql-deploy.yml
mohitrajain: yes, i will do this. Old habit of skipping things

ansible/postgresql-playbooks/postgresql-deploy-primary.yml
mohitrajain: Here, we need the IP address. We are already configuring the node groups postgresql_rw and postgresql_ro in hosts.ini. If we specify more vars specific to primary, replica_1 and replica_2 with their IP address, I found it overkill as hosts has been already grouped. Hence, this big expression to cover corner cases, if someone didn't specify group then what should be done.

ansible/helm_external.yml
mohitrajain: yes, let me fix it.

offline/tasks/pre_chart_process_0.sh
mohitrajain: will remove it

offline/postgresql-cluster.md
mohitrajain: let me fix it

offline/postgresql-cluster.md
mohitrajain: yes, makes sense, with tags and skip-tags. I will modify it and update the document

ansible/inventory/offline/group_vars/all/offline.yml
Veki301: thats okay, just a simple move for clarity

ansible/postgresql-playbooks/postgresql-deploy-primary.yml
Veki301: why not just settle right away, and say postgresql1 is primary, and postgresql2 and 3 are replica1 and 2, that structure/grouping is already in the hosts.ini, if somebody didnt input an IP for those nodes, just throw an error? anyways, I dont think we need to expect ambiguity here around IPs, which we mandate

mohitrajain: here, I didn't want to fix it, because later if we wish to change the structure like postgresql2 act as primary and postgresql1 acting as replica, then we can change the group instead of changing the IPs. Here the assumption is that, once postgresql[1-3] nodes are defined, we can simply change their groups if required, instead of changing their IPs. Working with groups would be more easy then working with IPs, hence this logic to ease out the pain of remembering which IPs belong to which group.

